PR #55, opened 2017-12-15
min/max peptide length comparison in vector visualization 

yang-yangfeng:
comparisons for peptide length should account for peptides exactly equal to the min/max length


susannasiebert:
CI, test this.

PR #57, opened 2017-12-18
Update integrate input file parser to use `can be inframe` column to determine fusion type

### LINKED ISSUE #53: pVACfuse ignore 3p tails of fusion peptides which can be in frame
as per discussion with @chrisamiller and @susannasiebert :

Relevant columns from integrate-neo bedpe:
"can be inframe" boolean (Col 13), fusion peptide (Col 14), Transcripts (Col 17)

Transcripts column really contains 3 parts:
5p transcripts; 3p in-frame transcripts; 3p frame-shift transcripts;

Currently, pVACfuse uses the information from the transcripts column to determine which 3p transcripts are in frame and which are frameshifts, and attempts to create different entries/rows for them and treats them differently. However, as we now understand it, what's really going on is we have 1 peptide sequence inferred from the RNA data, and the frame of the transcripts is semi-irrelevant as all this is saying is whether the 3p portion of the peptide reflects a canonical/annotated peptide sequence. 

What we really care about is whether there is some in-frame transcript that, when translated in the canonical/annotated frame, contains the 3p portion of the annotated fusion peptide (of course, we are not taking into account whether this transcript is actually even expressed in our sample - perhaps a future feature?). Conveniently, this information is encoded in the "can be inframe" boolean value in column 13. This value is 1 if there is an in-frame 3p transcript which would produce the predicted peptide sequence, rendering the tail sequence of the 3p portion of the peptide "known" or "not novel". These fusion peptides still produces novel epitopes, as the fusion junction itself is a novel event. As such, **if there is a 1 in column 13, we should simply treat the fusion peptide as we treat missense snvs in pvacseqs, picking a 2n+1 window centered at the fusion event and considering only neoepitopes which overlap the fusion.** However, **if there is a 0 in column 13, we can consider any neoepitopes that cross into the 3p portion of the fusion peptide.**

susannasiebert:
@yang-yangfeng @chrisamiller can you please review this PR and confirm that this is in line with what we discussed recently.

I'm still not 100% convinced that this is the way to determine whether a fusion is frameshift vs inframe but this PR changes the code so this determination is based on the `can be inframe` column only and treating all transcripts for an entry with `1` in that column as `inframe`, which is what was discussed last week.

Closes #53


susannasiebert:
Yes and yes. The output tsv gets inputted to the fasta generator, where frameshift fusions and inframe fusions are handled appropriately.

PR #91, opened 2018-03-05
Fixes to get `pvacseq generate_protein_fasta` completing without exits

andreyto:
Please see the commit messages.

### DIFF
--- a/lib/input_file_converter.py
+++ b/lib/input_file_converter.py
@@ -268,7 +268,7 @@ class VcfConverter(InputFileConverter):
                     gene_name = transcript['SYMBOL']
                     index = '%s.%s.%s.%s' % (gene_name, transcript_name, consequence, amino_acid_change_position)
                     if index in indexes:
-                        print("Warning: TSV index already exists: {}".format(index))
+                        sys.exit("Warning: TSV index already exists: {}".format(index))
                     else:
                         indexes.append(index)
                     ensembl_gene_id = transcript['Gene']


lib/input_file_converter.py
```diff
@@ -265,7 +265,7 @@ def execute(self):
                     gene_name = transcript['SYMBOL']
                     index = '%s.%s.%s.%s' % (gene_name, transcript_name, consequence, amino_acid_change_position)
                     if index in indexes:
-                        sys.exit("TSV index already exists")
+                        print("Warning: TSV index already exists: {}".format(index))
```
susannasiebert: We should not change this to just a warning. Downstream steps in the main pVACseq pipeline require that the TSV index be unique between the different variants so that we can use it as a unique identifier. Would you be able to share some example data where you're running into this error?

lib/fasta_generator.py
```diff
@@ -50,7 +50,9 @@ def determine_flanking_sequence_length(self, full_wildtype_sequence_length, pept
 
     def get_wildtype_subsequence(self, position, full_wildtype_sequence, wildtype_amino_acid_length, peptide_sequence_length, line):
         one_flanking_sequence_length = self.determine_flanking_sequence_length(len(full_wildtype_sequence), peptide_sequence_length, line)
-        peptide_sequence_length = 2 * one_flanking_sequence_length + wildtype_amino_acid_length
+        ##clip by wt sequence length, otherwise with deletions peptide_sequence_length may exceeds full wt sequence length,
+        ##and the code below tries extracting ranges beyond the wt sequence
+        peptide_sequence_length = min(2 * one_flanking_sequence_length + wildtype_amino_acid_length,len(full_wildtype_sequence))
```
susannasiebert: Can you please provide some test data for this change?

tools/pvacseq/main.py
```diff
@@ -79,7 +79,7 @@ def main():
     except AttributeError as e:
         parser.print_help()
         print("Error: No command specified")
-        sys.exit(-1)
+        raise
```
susannasiebert: This is a note to myself, but we should make this change for all the tools.

lib/fasta_generator.py
andreyto: I have added a test to the test harness. It runs with my change, and fails without it. The target length of 600 in the test case is contrived to exceed ORF length in your test `input.vcf`, but it my real application it was failing with target length 60
```
pvacseq generate_protein_fasta tests/test_data/generate_protein_fasta/input.vcf 600 tmp.fasta
```

lib/input_file_converter.py
andreyto: Susanna,
Unfortunately, I cannot share my input VCFs. I was running this on internal proprietary data. Mutect2 WES VCF. Only one or two entries per sample were generating this error, so I did not have any choice but to turn this into a warning.
I thought that the step in `generate_protein_fasta.py` was the last in the pipeline, with no downstream analysis afterwards? I need these peptides for building a DB for proteomics peptide identification.
The code in `generate_protein_fasta.py` appears to be largely copy/pasted from `generate_fasta.py`. The latter one is probably that is used by downstream steps? Maybe then have an option in `input_file_converter.py` that turns exit into a warning when called only from `generate_protein_fasta.py`?

susannasiebert: `tools/pvacseq/generate_protein_fasta.py` is a standalone script that just generates the peptide fasta and is invoked by the `pvacseq generate_protein_fasta` command. The full `pvacseq run` command generates the protein fasta, makes binding predictions, and filters the output. `lib/generate_fasta.py` is a remnant back from the pVAC-Seq repository before we moved the codebase to pVACtools and reorganized everything. The code now lives in the FastaGenerator class (`lib/fasta_generator.py`). `lib/generate_fasta.py` should actually be deleted. You will see that `generate_protein_fasta.py` is command is using this class.

With your change what would happen is that you are potentially losing peptides because the same index/identifier points to two different peptide sequences and the first one gets overwritten in the `parse_files` method of `generate_protein_fasta.py`.

Would it be possible to come up with dummy data that recreates this error? The only situation I can think of where this might be happening is if there are two variants in your VCF with the same position that also have the same consequence. In that case I think your solution would work because the resulting peptide sequences between the two variants would be the same but I would want to make sure that we aren't creating a scenario where we are losing peptides.

andreyto: Susanna,
Attached if a single variant from a model cell culture that triggers the "duplicated index" condition. This single-variant file is enough:
```
$ pvacseq generate_protein_fasta scratch/dup_index.vcf 60 scratch/tmp.fasta
Converting VCF to TSV
Warning: TSV index already exists: CD34.ENST00000367036.missense.28S/F. Entry is: Record(CHROM=1, POS=208063100, REF=G, ALT=[A, GA])
Completed
Generating Variant Peptide FASTA and Key File
Completed
Parsing the Variant Peptide FASTA and Key File
Completed
```
Also attached is the same variant before VEP annotation, to see what has been added by VEP. It looks like the VEP annotation contain multiple transcripts triggering the condition.
I will leave it up to you to modify my changes as you see fit.
Thanks,
Andrey

[duplicated_index_test_data.zip](https://github.com/griffithlab/pVACtools/files/1782494/duplicated_index_test_data.zip)

susannasiebert: The bug that you were seeing with your example is that pVACseq is interpreting the VEP allele incorrectly for this case. Your example has two alternate alleles, `A` and `GA` with a `G` reference with corresponding alleles in the VEP CSQ field. However, if your entry only had the `GA` alternate allele, VEP would "normalize" this as an `A` allele, which is what pVACseq is looking at. Since in your example `A` actually exists in the CSQ field, pVACseq looks at the `A` VEP entries for both alternate alleles. I will figure out a way to handle this case.

susannasiebert:
@andreyto I believe one of our comment chains disappeared after commit 02d801f. To recap, the duplicate index error you were seeing was because pVACseq was interpreting the VEP allele incorrectly for the test case you provided. Your example has two alternate alleles, `A` and `GA` with a `G` reference with corresponding alleles in the VEP CSQ field. However, if your entry only had the `GA` alternate allele, VEP would "normalize" this as an `A` allele, which is what pVACseq is looking at. Since in your example `A` actually exists in the CSQ field, pVACseq looks at the `A` VEP entries for both alternate alleles.

This bug was fixed in commit 5d103a9 and your example added as a test case. I also put the exit condition back. Feel free to pull my changes and test it on your full VCF but I believe that this PR is ready to be merged.

Thanks again for this pull request. It is much appreciated!

susannasiebert:
Merged. I will try to make a new release either today or tomorrow.

Issue #95, opened 2018-03-14
Create a priority rank


susannasiebert:

susannasiebert:
This has been implemented in PR #113  (merged). It will become available in version 1.1.

PR #113, opened 2018-04-17
Improvements and additions to report creation

### LINKED ISSUE #94: Output a condensed report

susannasiebert:
This PR does a few things:
- It changes how we group epitopes for the top score filter. With this change a "mutation" is determined by the chromosome, start, stop, ref and alt. Entries with the same set of those are treated as the same mutation and the top score filter is applied to that set.
- The top score filter is now run every time. As a result, the `--top-score-per-mutation` flag has been removed.
- If both class I and class II are run, we are now creating a set of combined reports in a subfolder called `combined`. The combined report is an aggregate of `MHC_Class_I/<sample_name>.combined.parsed.tsv` and `MHC_Class_II/<sample_name>.combined.parsed.tsv`. All filters are applied on the combined file.
- From the <sample_name>final.tsv a condensed report (`<sample_name>.final.condensed.tsv`) is created based on the specs provided by @chrisamiller. This is done for class I, class II, and the combined final report.
- From the condensed report(s) a ranked report (`<sample_name>.final.condensed.ranked.tsv`) is created based on the criteria provided by @chrisamiller. This is done for class I, class II, and the combined set.

TO-DO: add documentation for the new files as well as some explanation about how we condense the report and rank the epitopes. Also, update the example data folders.

Closes #94 and #95


lib/rank_epitopes.py
```diff
@@ -0,0 +1,35 @@
...
+            'Score',
+        ]
+
+    def execute(self):
+        df = pd.read_csv(self.input_file, sep='\t', index_col=False)
+        df['mt_score_rank'] = df['MT IC50'].rank(numeric_only=True, ascending=False, method='dense').fillna(value=0.0)
+        df['fold_change_rank'] = df['Fold Change'].rank(numeric_only=True, ascending=True, method='dense').fillna(value=0.0)
+        df['mt_allele_exp'] = df['Tumor RNA VAF'] * df['Gene Expression']
+        df['mt_allele_exp_rank'] = df['mt_allele_exp'].rank(numeric_only=True, ascending=True, method='dense').fillna(value=0.0)
+        df['tumor_dna_vaf_rank'] = df['Tumor DNA VAF'].rank(numeric_only=True, ascending=False, method='dense').fillna(value=0.0)
```
chrisamiller: Is this line right?  I expect that the above two lines ('mt_allele_exp' and 'Tumor DNA VAF') would both be ranked the same way, since higher scores are better for either.   The difference in "ascending" values makes me think that one of them isn't right.

chrisamiller:
Other than the possible ranking issue above, I don't see any issues here

lib/rank_epitopes.py
susannasiebert: You are correct that this is the wrong order. I think I got Tumor DNA VAF mixed up with normal VAF in my head. Good catch. I will fix it tomorrow. 

Btw, `numeric_only=True` Means that NA values do not get ranked and the `fillna` Then sets NAs to 0.0. `method` defines how equal values get ranked. `dense` will assign them all the same rank and then ranks increase by 1 between groups. This seemed like the most reasonable option to me but you might want to check the documentation (https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.rank.html) to see if one of the other ones would be a better choice.

Issue #111, opened 2018-04-12
Duplicated lines in output


ktroule:
Hi, once more.
I've few question related to the output of pVACseq. Sorry if these question are already solved (I've not found them).

1º As far as I know the VEP DownstreamProtein plugin only annotates predictions for transcripts suffering of frameshfit mutations. How do you calculate the the protein change when a SNV is happening? I would be interested in having the whole protein prediction, not only when a frameshift happens but also when there is a SNV. I guess you must be doing this in pVACseq as the output contains the MT Epitope Seq which is needed for the affinity prediction.

2º See image. I've highlighted transcripts (same color, same transcripts) .
Why are there repeated lines (2 and 8, 3 and 9, 4 and 10, and so on), the rest of column sshare same parameters between those lines (scores, etc). If I run several times the same pVACseq script, Does the program overwrites the final tsv or it just adds new lines? If it appends new runs of pvacseq that might be the answer.
![pvac_out](https://user-images.githubusercontent.com/9655951/38679740-c6a92894-3e64-11e8-9548-f2f6b316fbe5.png)

Once more, thanks for your patience and help-

ktroule:
Thanks! You can close the thread.

ktroule:
I've been trying to use  `pvacseq generate_protein_fasta`, it runs ok with my own VCF files, but I've also tried to run it using a Cosmic vcf file. I've annotated this Cosmic vcf using the exactly same VEP parameters I use for my own vcf. But I get this error:
```
Converting VCF to TSV
Traceback (most recent call last):
  File "/usr/local/bin/pvacseq", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.5/site-packages/tools/pvacseq/main.py", line 78, in main
    args[0].func.main(args[1])
  File "/usr/local/lib/python3.5/site-packages/tools/pvacseq/generate_protein_fasta.py", line 111, in main
    convert_vcf(args.input_file, temp_dir)
  File "/usr/local/lib/python3.5/site-packages/tools/pvacseq/generate_protein_fasta.py", line 52, in convert_vcf
    converter.execute()
  File "/usr/local/lib/python3.5/site-packages/lib/input_file_converter.py", line 202, in execute
    genotype = entry.genotype(vcf_reader.samples[0])
IndexError: list index out of range
```

Does this program take any argument from the header that I might be missing from the Cosmic vcf file?

susannasiebert:
[...]

PR #114, opened 2018-04-18
Handle variants with multiple stops in the amino acid change

susannasiebert:
This should fix the latest error in #111.


chrisamiller:
+1  Looks good to me!

PR #130, opened 2018-06-01
Remove top score per mutation from API

### LINKED ISSUE #123: Remove the `--top-score-per-mutation` flag
https://github.com/griffithlab/pVACtools/pull/113 Removed the `--top-score-per-mutation` flag (the top score filter now gets run by default). This option needs to be removed from the API/client.

@atwollam this is a pretty straight-forward change so I think this is good issue to tackle for your first PR.

atwollam:
Resolves #123 by adjusting the API to account for the top score filter being run by default.

### DIFF
--- a/lib/pipeline.py
+++ b/lib/pipeline.py
@@ -290,6 +290,11 @@ class Pipeline(metaclass=ABCMeta):
     def top_result_filter_out_path(self):
         return os.path.join(self.output_dir, self.sample_name+".filtered.top.tsv")
 
+    def top_result_filter(self):
+        status_message("Running Top Score Filter")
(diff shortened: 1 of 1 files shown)


susannasiebert:
It looks like you are trying to merge into master instead of staging. We use the staging branch to consolidate all feature additions before a release. You can change the base branch for this PR by clicking the `Edit` button next to the title.

Some of the changes you made in this PR where already present/addressed in a previous PR that was merged into staging. So you will probably see some merge conflicts that you will need to address.

Also, the integration tests are currently failing but that might resolve itself once you change the base branch to staging.

PR #135, opened 2018-06-15
Add MHCnuggets prediction algorithm

### LINKED ISSUE #67: Investigate using MHCnuggets as a prediction algorithm
Original issue: https://github.com/griffithlab/pVAC-Seq/issues/321

https://github.com/KarchinLab/mhcnuggets

susannasiebert:
The current mhcnuggets package is not python3 compatible but a PR with the necessary changes has been merged into master on the MHCnuggets-2.0 repository. I added a line in the .travis.yml to install mhcnuggets from master, which can be taken out once a new package with the changes has been released.

The allele format required my MHCnuggets is quite different from IEDB so if someone wants to run allele HLA-A*02:01 on NetMHC and MHCnuggets they would need to specify this allele twice, once in the format `HLA-A*02:01` (for NetMHC) and once as HLA-A0201 (MHCnugget). We should consider defining a standardized set of alleles for pVACtools and handling the conversion internally.

Closes #67

### DIFF
--- a/tools/pvacseq/run.py
+++ b/tools/pvacseq/run.py
@@ -166,9 +166,17 @@ def main(args_input = sys.argv[1:]):
         if allele in MHCI.all_valid_allele_names():
             class_i_alleles.append(allele)
             valid = 1
+        elif 'MHCnuggetsI' in args.prediction_algorithms:
+            if allele.startswith(("HLA-A", "HLA-B", "HLA-C")):
+                class_i_alleles.append(allele)
+                valid = 1
         if allele in MHCII.all_valid_allele_names():
             class_ii_alleles.append(allele)
             valid = 1
+        elif 'MHCnuggetsII' in args.prediction_algorithms:
+            if allele.startswith(("DP", "DQ", "DO", "DM", "DR", "H2")):
+                class_ii_alleles.append(allele)
+                valid = 1
         if not valid:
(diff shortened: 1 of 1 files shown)


huang-ying:
I ran following command
```bash
pvacfuse run pvacfuse_example_data/fusions.bedpe.annot Test HLA-A0201 MHCnuggetsI pvacfuse_example_data/mhcnuggets -e 9
```
and got following error message
```
Building model
Invalid human allele
Closest allele found None
```
It seems mhcnuggets requires HLA-A02:01, change it in `pVACtools/tools/pvacseq/iedb_alleles/class_i/MHCnuggets.txt` works

susannasiebert:
Yes, we are aware of this issue, which is why this PR hasn't been merged yet. We're in contact with the folks that created MHCnuggets to finalize the HLA naming scheme. Please see https://github.com/KarchinLab/mhcnuggets-2.0/issues/2 for more information on that progress.

susannasiebert:
I made some changes to this implementation to do some rudimentary checking of an allele name and see if it matches the HLA format. This would allow us to provide any possible HLA allele to MHCnuggets. However, with this change only human (HLA) alleles can be used with MHCnuggets. The format checking also expects IEDB-formatted HLAs and transposes them to the MHCnuggets format (which is still slightly different) but this has the advantage that people can use the same HLA names consistently.

PR #141, opened 2018-06-28
config files generated for runs provided in dropbox

### LINKED ISSUE #126: When existing run gets dropped in the dropbox folder, create the config.json file
The standalone pVACseq runs create their own log files in `MHC_Class_I/log/inputs.yml` and `MHC_Class_II/log/inputs.yml` (depending on what class of alleles and prediction algorithms are run). If the results folder of such a standalone run gets dropped into the dropbox folder, the config.json file doesn't exist. In that case, create it from the existing inputs.yml file(s). If a file exists for both class I and class II, you'd need to combine them. Only the list of alleles and prediction algorithms should differ and those just each need to be combined.

atwollam:
Resolves #126 by adding to the api the ability to generate config.json files for runs added to the dropbox

### DIFF
--- a/utils/pvacapi/controllers/utils.py
+++ b/utils/pvacapi/controllers/utils.py
@@ -307,6 +307,11 @@ def initialize(current_app, args):
                                     MHC_dict.update({k:v for k,v in temp_dict.items() if k not in MHC_dict})
                                     MHC_dict['alleles'].extend(temp_dict['alleles'])
                                     MHC_dict['prediction_algorithms'].extend(temp_dict['prediction_algorithms'])
+                        del MHC_dict['tmp_dir']
+                        MHC_dict['output'] = MHC_dict['output_dir']
+                        del MHC_dict['output_dir']
+                        if 'MHC_Class' in os.path.basename(MHC_dict['output']):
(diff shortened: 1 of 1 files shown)


susannasiebert:
The API tests are failing after this change:
```
FAIL: test_full_api_pipeline (tests.test_api.APITests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/griffithlab/pVACtools/tests/test_api.py", line 610, in test_full_api_pipeline
    self.assertTrue(finaltsv)
AssertionError: [] is not true
```

susannasiebert:
I was able to confirm that the `config.json` file gets created. A few things that need to be adjusted the bring the contents more in line with the pvacapi-created `config.json` file:
- `output_dir` needs to be adjusted: In my example it is `/home/archive/ssiebert/pVAC-Seq.new/pvacseq/example_data/MHC_Class_I`. We'd want it to be `/home/archive/ssiebert/pVAC-Seq.new/pvacseq/example_data`. So basically, when you write this `output_dir` you need to manipulate the value to use the parent directory, which is what pvacapi records this as.
- `tmp_dir`: this doesn't exist in the pvacapi-created file. This should just be skipped.

PR #154, opened 2018-07-17
added pvacviz entrypoint and basic web server to host client

jmcmichael:
This PR adds a new entrypoint to pvactools, pvacviz, an app that implements a very basic webserver to host the pVACviz client's compiled javascript, css, images, etc. Note that the compiled client isn't included in the repo, instead they'll be added to the pvacviz/client directory and included as part of the pypi package deployment.

Also renamed the pacseq-api endpoint to pvacapi.

### DIFF
--- a/utils/pvacapi/visualizations.py
+++ b/utils/pvacapi/visualizations.py
@@ -145,7 +145,7 @@ x_field = Select(
     options=sorted([
         (key, val) for (key, val) in cols.items()
     ], key = lambda x:x[1]),
-    value = 'corresponding_wt_score'
+    value = 'corresponding_wt_score' if 'corresponding_wt_score' in cols.keys() else 'wt_ic50'
 )
 widgets.append(x_field)
 y_field = Select(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -43,17 +43,37 @@
                 filename
             ))
 
+for dirpath, dirnames, filenames in os.walk("utils/pvacviz/client"):
+    for filename in filenames:
+        if not (filename.endswith(".py") or filename.endswith(".pyc")):
+            server_data.append(os.path.join(
+                os.path.relpath(
+                    dirpath,
+                    'utils/pvacapi'
```
susannasiebert: Change this to `utils/pvacviz/client`.

setup.py
```diff
@@ -43,17 +43,37 @@
                 filename
             ))
 
+for dirpath, dirnames, filenames in os.walk("utils/pvacviz/client"):
```
susannasiebert: Add a new array `client_data = []` before this line

setup.py
```diff
@@ -43,17 +43,37 @@
...
+    for filename in filenames:
+        if not (filename.endswith(".py") or filename.endswith(".pyc")):
+            server_data.append(os.path.join(
```
susannasiebert: Append to the `client_data` array instead of `server_data`.

susannasiebert:
I tested your newest commits and it's working for me. Very cool. I noticed that it opens the manage page, not the start page. Is there a reason for one vs the other?

jmcmichael:
Yeah I chose manage as the default, figuring that it would be the most common page user would want to see.

PR #162, opened 2018-08-02
Reworked the result file workflow/lifecycle

### LINKED ISSUE #146: Rework the result file workflow/lifecycle
After discussing with Josh we came up with the follow workflow for how results are handled:

- The folder currently named `/results` will be renamed `/.processes`. This will make it hidden from the user. We want to prevent users from interacting with this folder directly since this is where results are directly written to. This will prevent buggy behavior caused by users inadvertently moving files around.
- We will have three actions for a result (these apply to processes that are not currently running):
1) `Archive`. This will **move** the result to the `/archive` folder and remove the process from the process list.
2) `Export`. This will **copy** the result to the `/export` folder but keep the process and its result in the process list
3) `Delete`. This will permanently **remove** the results files and the process from the process list
- If a user wants to visualize an existing result - one that was either created form the interface and archived, or that was created from the command line - they will copy that result directory to the `/visualize` folder (formally named `/dropbox`).

atwollam:
Resolves #146 by adding in export and delete actions and renaming files appropriately.

### DIFF
--- a/utils/pvacapi/config/swagger.yaml
+++ b/utils/pvacapi/config/swagger.yaml
@@ -1057,8 +1057,8 @@ paths:
           description: "Unexpected error"
           schema:
             $ref: "#/definitions/Error"
-  /api/v1/delete/{processID}:
-    get:
+  /api/v1/processes/{processID}/delete:
+    delete:
       tags:
         - "delete"
       summary: "delete a process"
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -665,8 +665,8 @@ class APITests(unittest.TestCase):
             )
             self.assertEqual(response.status_code, 200, response.url+' : '+response.content.decode())
             self.assertIsInstance(response.json(), dict)
-        response = requests.get(
-            self.urlBase+'/delete/%d'%processID,
+        response = requests.delete(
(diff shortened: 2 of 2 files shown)


utils/pvacapi/config/swagger.yaml
```diff
@@ -1035,6 +1035,50 @@ paths:
           description: "Unexpected error"
           schema:
             $ref: "#/definitions/Error"
+  /api/v1/export/{processID}:
```
susannasiebert: In order to make this conform to REST better, this should be `/api/v1/process/{processID}/export` and should be a POST request instead of GET.

utils/pvacapi/config/swagger.yaml
```diff
@@ -1035,6 +1035,50 @@ paths:
...
+      responses:
+        200:
+          description: "No meaningful data to return"
+          schema:
+            type: "string"
+        default:
+          description: "Unexpected error"
+          schema:
+            $ref: "#/definitions/Error"
+  /api/v1/delete/{processID}:
```
susannasiebert: Analogous to above, this should be `/api/v1/process/{processID}/archive` and should be a DELETE request instead of GET.

PR #183, opened 2018-08-27
Decode hgvs strings to remove hex encoding of special characters

susannasiebert:


malachig:
Awesome.  Thanks @susannasiebert

PR #188, opened 2018-09-04
 Add field `is_input` to input file objects

### LINKED ISSUE #167: inputs API endpoint should return a field is_input for each file object
This will be set to `True` if the file ends on `.vcf` or `.vcf.gz`.

susannasiebert:
Closes #167


jmcmichael:
ok to merge - pvacviz input service now filters on is_input

PR #189, opened 2018-09-04
Change the restart endpoint to be POST to /api/v1/processes/{processID}/restart instead of GET to /api/v1/restart/{processID}

susannasiebert:


jmcmichael:
pvacviz client has been updated to handle these new endpoints, OK to merge.

PR #201, opened 2018-09-21
Add parsing and filtering on transcript support level to pVACseq

### LINKED ISSUE #191: Add TSL-support to pVACseq
This would include:
- If the VCF was annotated with the `--tsl` option, we pull out the TSL into the all_epitopes and filtered report. It does not get added to the condensed reports.
- Add a filter to be able to filter on the TSL option. The default is TSL 1. This filter will be run by default if TSL is in the VCF.

susannasiebert:
The new transcript_support_level_filter gets run by default as part of the pVACseq pipeline with a TSL cutoff of <=1 which can be overwritten using the `--maximum-transcript-support-level` parameter. By default, NA values (which means, either the TSL is NA or no TSL annotations were found in the VCF), will not be filtered out. This behavior is consistent with all other filters and can be overwritten using the `--exclude-NAs` flag.

Closes #191

### DIFF
--- a/lib/pipeline.py
+++ b/lib/pipeline.py
@@ -54,6 +54,7 @@ class Pipeline(metaclass=ABCMeta):
         self.trna_cov                    = kwargs.pop('trna_cov', None)
         self.trna_vaf                    = kwargs.pop('trna_vaf', None)
         self.expn_val                    = kwargs.pop('expn_val', None)
+        self.maximum_transcript_support_level = kwargs.pop('maximum_transcript_support_level', None)
         self.additional_report_columns   = kwargs.pop('additional_report_columns', None)
         self.fasta_size                  = kwargs.pop('fasta_size', 200)
         self.iedb_retries                = kwargs.pop('iedb_retries', 5)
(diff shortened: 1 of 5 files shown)


tools/pvacseq/transcript_support_level_filter.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    parser.add_argument(
+        "--maximum-transcript-support-level", type=int,
+        help="The threshold to use for filtering epitopes on the transcript support level. "
+        +"Keep all epitopes with a transcript support level <= to this cutoff. Default: 1",
+        default=1,
+        choices=[1,2,3,4,5]
+    )
+    parser.add_argument(
+        '--exclude-NAs',
```
chrisamiller: 1) I poked around the code, but I don't see where this option is exposed to the end user (through, say, "pvacseq run").  (I probably just missed it, though)
2) It looks like the "exclude-NAs" argument applies to all filters, so there's no way to exclude NAs from filter X, but keep them in filter Y. Is that correct?  (and the desired behavior?)  I'm not sure - thinking about it...

susannasiebert: 1. It's exposed in lib/run_argument_parser.py line 129 (https://github.com/griffithlab/pVACtools/pull/201/files#diff-a06dc88b532df643a77a13749169da4aR129). It already existed before this PR so that's why it's not showing in the diffs
2. Correct. I'm not sure if there is a good way around this. A user can always filter manually on the all_epitopes file later, if desired. This new filter is also available standalone for that purpose.

chrisamiller: Works for me - +1!

PR #205, opened 2018-09-25
added maximum transcript support level field to API & relevant functions

jmcmichael:


jmcmichael:
The help text likely needs to be updated to something more helpful, I just included the field name as a placeholder.

susannasiebert:
@jmcmichael on the command line the help text is:

PR #208, opened 2018-10-02
Some copyediting suggestions for the docs.

tmooney:
I saw some new commits on #128 and ended up reading through most of the changes.  A few of these definitely have alternative ways they could be adjusted, and a few comma changes are fairly opinionated :smile:


jmcmichael:
+1

PR #211, opened 2018-10-08
Update data type of vaf arguments to float

### LINKED ISSUE #210: Manual VAF filters 'invalid int value'
* pvactools version: 1.1
* Python version: 3.5
* Operating System: High Sierra

**Describe the bug**
Setting manual VAF cutoffs in the pvacseq run command not working. 

**To Reproduce**
pvacseq run -e 8,9,10,11 --fasta-size 100 --downstream-sequence-length 500 --pass-only --binding-threshold 500 --iedb-retries 50 --tdna-cov 5 **--tdna-vaf 0.05** --iedb-install-directory /opt/iedb -i $Yaml_file --normal-sample-name NORMAL $Input_vcf TUMOR HLA-A*01:01,HLA-A*24:02,HLA-B*07:02,HLA-B*37:01,HLA-C*07:02,HLA-C*06:02 MHCnuggetsI NNalign NetMHC NetMHCIIpan NetMHCcons NetMHCpan PickPocket SMM SMMPMBEC SMMalign $Output_dir

"pvacseq run: error: argument --tdna-vaf: invalid int value: '0.05'"

susannasiebert:
Closes #210

### DIFF
--- a/lib/run_argument_parser.py
+++ b/lib/run_argument_parser.py
@@ -61,7 +61,7 @@ class RunArgumentParser(metaclass=ABCMeta):
         parser.add_argument(
             "-r", "--iedb-retries",type=int,
             default=5,
-            help="Number of retries when making requests to the IEDB RESTful web interface. Must be less than or equal to 100."
+            help="Number of retries when making requests to the IEDB RESTful web interface. Must be less than or equal to 100. "
                  + "Default: 5"
         )
         parser.add_argument(
--- a/tools/pvacseq/coverage_filter.py
+++ b/tools/pvacseq/coverage_filter.py
@@ -53,7 +53,7 @@ def define_parser():
     )
     parser.add_argument(
         '--expn-val', type=int,
-        help="Gene and Transcript Expression cutoff. Sites above this cutoff will be considered"
+        help="Gene and Transcript Expression cutoff. Sites above this cutoff will be considered. "
              + "Default: 1",
         default=1
     )


lib/run_argument_parser.py
```diff
@@ -177,19 +177,19 @@ def __init__(self):
             default=10
         )
         self.parser.add_argument(
-            '--normal-vaf', type=int,
+            '--normal-vaf', type=float,
             help="Normal VAF Cutoff. Sites BELOW this cutoff in normal will be considered. " +
             "Default: 0.02",
             default=0.2
```
tmooney: The default and the documented value don't match.

tmooney: (Seems like it ought to document the defaults for you...)

susannasiebert:
I know, right? That seems dumb. It is illustrated in this part of the docs (https://docs.python.org/3/library/argparse.html#default):
```
If the default value is a string, the parser parses the value as if it were a command-line argument. In particular, the parser applies any type conversion argument, if provided, before setting the attribute on the Namespace return value. Otherwise, the parser uses the value as is:

>>>
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('--length', default='10', type=int)
>>> parser.add_argument('--width', default=10.5, type=int)
>>> parser.parse_args()
Namespace(length=10, width=10.5)
```

According to the above maybe all of our defaults should be strings so that they get type-cast.

lib/run_argument_parser.py
susannasiebert: Good catch. I fixed that. Unfortunately, the defaults don't appear when doing `--help`. Apparently this can be changed by setting the `formatter_class` argument of the `add_argument` to `argparse.ArgumentDefaultsHelpFormatter`.

This seems reasonable so I will make an issue to do that.

PR #224, opened 2018-11-26
Support empty transcript and gene expression fields

susannasiebert:


susannasiebert:
@tmooney Once we remove the option to input additional input files I replan on refactoring this portion and removing this code duplication.

PR #227, opened 2018-11-30
Pvacvector visualization

susannasiebert:
The package to run `covert` is difficult to set up. Leaving the vaccine vector graphic as a `.ps` file seems sufficient.

This also increases the canvas width to 1200 pixels which is needed for the longer epitopes names.

### DIFF
--- a/lib/vector_visualization.py
+++ b/lib/vector_visualization.py
@@ -1,6 +1,7 @@
 import turtle
 import os
 import sys
+from PIL import Image
 
 class VectorVisualization:
     def __init__(self, input_fasta, output_directory):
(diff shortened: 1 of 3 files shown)


chrisamiller:
Thinking ahead, will that ps file display in browser? (assuming that we'll wrap this into the web interface eventually)

susannasiebert:
Good point. It does not display. I will work on an alternative. For now I think this is a good stop-gap measure to get this running more reliably.

chrisamiller:
Yeah, I have no objection to the change in general! I just know that lots of people will ultimately want a jpg/gif/png that's less complicated to view

susannasiebert:
Good new everyone! I found a python module to convert images without having to make a system call to a command line tool. I pushed a commit to use that instead.

chrisamiller:
Nice!  +1

PR #244, opened 2019-01-10
Fix that pVACvector was not fully eliminating peptide-spacer-peptide combinations that contained well-binding epitopes

susannasiebert:
pVACvector now eliminates any peptide-spacer-peptide combination that contains one or more epitopes that are good binders. 
If there are no peptide-spacer-peptide combinations remaining for peptide1 to peptide2, then the edge from peptide1 to peptide2 will not not be added to the graph, i.e. all of the combinations produce well-binding epitopes.
If the annealing procedure attempts to traverse a path that contains one of these non-existing edges, the length for this path is returned as 0 so that it is moves to the bottom of the options. We're looking for the longest path and we don't want to consider paths that only contain peptide-spacer-peptide options with well-binding epitopes. 
After creating the path but before running the annealing procedure we check that all nodes have at least one outgoing edge and throw a fatal error if we find a node without outgoing edges because otherwise we wouldn't be able to find a cyclical path, which is what is desired.


chrisamiller:
> we check that all nodes have at least one outgoing edge and throw a fatal error if we find a node without outgoing edges because otherwise we wouldn't be able to find a cyclical path

This is close but not completely accurate. A cyclical path isn't necessary, because there will be some stuff in the vector besides the peptides (a promotor site and spacer stuff that's used to close the vector). Finding a linear path is sufficient. It may be that when the algorithm was set up the first time, this wasn't completely clear to those who wrote it.

tools/pvacvector/run.py
```diff
@@ -105,54 +105,66 @@ def run_pipelines(input_file, base_output_dir, args):
...
     for ep in seq_tuples:
         ID_1 = ep[0]
         ID_2 = ep[1]
-        Paths.add_node(ID_1)
-        Paths.add_node(ID_2)
         for space in spacers:
             if space is None:
                 key = str(ID_1 + "|" + ID_2)
             else:
                 key = str(ID_1 + "|" + space + "|" + ID_2)
```
chrisamiller: totally minor point - might be better to use "for spacer in spacers", as I briefly thought this was a literal space character for some reason

tools/pvacvector/run.py
```diff
@@ -202,17 +238,19 @@ def find_optimal_path(Paths, distance_matrix, seq_dict, seq_keys, base_output_di
...
+                min_score = min(min_score, edge['weight'])
+                cumulative_weight += edge['weight']
+                all_scores.append(str(edge['weight']))
+                spacer = edge['spacer']
                 if spacer is not '':
                     name.append(spacer)
-            except IndexError:
-                continue
+            else:
+                sys.exit("Unable to find path. All possible peptides for edge '{} - spacer - {}' contain at least one eptiope that is a good binder.".format(state[i], state[i + 1]))
```
chrisamiller: spelling of "epitope"

tools/pvacvector/run.py
```diff
@@ -105,69 +105,107 @@ def run_pipelines(input_file, base_output_dir, args):
...
                 else:
                     Paths.add_edge(ID_1, ID_2, weight=worst_case, spacer='')
 
     print("Graph contains " + str(len(Paths)) + " nodes and " + str(Paths.size()) + " edges.")
     return Paths
 
+def check_graph_valid(Paths):
+    error_text = ('A vaccine design using the parameters specified could not be found.  Some options that you may want to consider:\n' +
+                 '1) increasing the acceptable junction binding score to allow more possible connections (-b parameter)\n' +
+                 '2) using the "median" binding score instead of the "best" binding score for each junction, (best may be too conservative, -m parameter)')
```
chrisamiller: This option best/median doesn't exist yet (but should be exposed!)

susannasiebert:
@chrisamiller I just pushed a few commits to address your comments.

chrisamiller:
I haven't tried to run the changes or anything, but don't see any other issues!  +1

PR #245, opened 2019-01-14
Parallelize the calls to IEDB

susannasiebert:
This PR uses the pymp module to parallelize the calls to IEDB. It
parallelizes on the chunked input fasta so that file chunks can be
processed in parallel to each other.

We could attempt to further parallelize by applying the same logic to the allele/prediction algorithm/epitope length loops but since each of those calls would attempt to read the same file chunk, I wasn't sure whether that would lead to problems.


susannasiebert:
File changes are best viewed including `&w=1` in the URL to ignore whitespace changes.

chrisamiller:
+1 - this looks nice

malachig:
+1 agree, this is going to be fantastic.

susannasiebert:
I just pushed a commit that also parallelizes the other for-loops in this function and a method that attempts to load balance all the nested loops.

lib/pipeline.py
```diff
@@ -248,6 +250,28 @@ def generate_fasta(self):
...
 
+    def balance_multithreads(self, iteration_info):
+        for i in range(1,self.n_threads):
+            #find the dimension with the most iterations per thread
+            (max_iterations, dimension) = max(zip(map(lambda x: x['iterations_per_thread'], iteration_info.values()), iteration_info.keys()))
+            #assign that dimension one more thread
+            iteration_info[dimension]['threads'] += 1
+            #calculate how many total threads this combination would use since these are nested threads
+            total_n_threads = iteration_info['file']['threads'] * iteration_info['allele']['threads'] * iteration_info['length']['threads'] * iteration_info['algorithm']['threads']
+            #if total_n_threads exceeds the total number of threads reqeusted, reset to previous state and return
```
tmooney:
```suggestion
            #if total_n_threads exceeds the total number of threads requested, reset to previous state and return
```

lib/pipeline.py
```diff
@@ -248,6 +250,28 @@ def generate_fasta(self):
...
+            if total_n_threads > self.n_threads:
+                iteration_info[dimension]['threads'] -= 1
+                return iteration_info
+            #recalculcate how many iteration per thread that dimension will have with that new number of threads
```
tmooney:
```suggestion
            #recalculate how many iterations per thread that dimension will have with the new number of threads
```

lib/pipeline.py
```diff
@@ -248,6 +250,28 @@ def generate_fasta(self):
...
+            #find the dimension with the most iterations per thread
+            (max_iterations, dimension) = max(zip(map(lambda x: x['iterations_per_thread'], iteration_info.values()), iteration_info.keys()))
+            #assign that dimension one more thread
+            iteration_info[dimension]['threads'] += 1
+            #calculate how many total threads this combination would use since these are nested threads
+            total_n_threads = iteration_info['file']['threads'] * iteration_info['allele']['threads'] * iteration_info['length']['threads'] * iteration_info['algorithm']['threads']
+            #if total_n_threads exceeds the total number of threads requested, reset to previous state and return
+            if total_n_threads > self.n_threads:
+                iteration_info[dimension]['threads'] -= 1
+                return iteration_info
```
tmooney: Is it possible that adding a thread to one of the other dimensions could be useful and still end up under `n_threads`?

(I was thinking about this using 2, 3, 5, and 8.  `(30)*8 => (30)*9` yields `270` but `2*(120) => 3*(120)` yields `360`.  Not that anyone is running with 300 threads...)

susannasiebert: Oh yeah, absolutely. I couldn't think of a good way to test though these options without it getting too complicated.

PR #246, opened 2019-01-14
Fix typos: s/migt/might/; default is actually 25.

tmooney:
I really wish this would print the defaults without them being duplicated in the documentation string!


susannasiebert:
Sounds like it's time to merge in https://github.com/griffithlab/pVACtools/pull/213

PR #269, opened 2019-01-24
Make spacers an optional parameter in pVACvector

### LINKED ISSUE #261: pVACvector - Expose spacers as a parameter
We should expose the list of vaccine spacers as a (csv) parameter. 
- The default should be all of the current list
- there needs to be someway to specify "no spacers" (empty string? flag?)

For one this would allow users to change that list if they have inside knowledge. For another, this could potentially help a lot with runtime as well - a user could start with no spacers, then if no solution is found, could slowly increase the number.  There is probably additional room for more intelligent use of spacers, but this would be a great start!

susannasiebert:
Closes #261

### DIFF
--- a/lib/run_argument_parser.py
+++ b/lib/run_argument_parser.py
@@ -231,6 +231,6 @@ class PvacvectorRunArgumentParser(RunArgumentParser):
         )
         self.parser.add_argument(
             '--spacers', type=lambda s:[spacer for spacer in s.split(',')],
-            help="Comma-separated list of spacers to use for testing junction epitopes. Include an empty string (\"\") to test junctions without spacers.",
-            default="\"\",HH,HHC,HHH,HHHD,HHHC,AAY,HHHH,HHAA,HHL,AAL"
+            help="Comma-separated list of spacers to use for testing junction epitopes. Include None to test junctions without spacers.",
+            default="None,HH,HHC,HHH,HHHD,HHHC,AAY,HHHH,HHAA,HHL,AAL"
         )
--- a/lib/fasta_generator.py
+++ b/lib/fasta_generator.py
@@ -351,7 +351,7 @@ class VectorFastaGenerator():
                 trunc_seq2 = seq_dict[seq2][0:wingspan_length]
 
                 for this_spacer in self.spacers:
-                    if this_spacer != '""':
+                    if this_spacer != 'None':
                         seq_ID = seq1 + "|" + this_spacer + "|" + seq2
                         epitopes[seq_ID] = (trunc_seq1 + this_spacer + trunc_seq2)
                     else:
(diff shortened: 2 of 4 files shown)


lib/run_argument_parser.py
```diff
@@ -229,3 +229,8 @@ def __init__(self):
             '-n', "--input-n-mer", default='25',
             help="Length of the peptide sequence to use when creating the FASTA from the pVACseq TSV.",
         )
+        self.parser.add_argument(
+            '--spacers', type=lambda s:[spacer for spacer in s.split(',')],
+            help="Comma-separated list of spacers to use for testing junction epitopes. Include an empty string (\"\") to test junctions without a spacer.",
```
chrisamiller: s/without a spacer/without spacers/

lib/run_argument_parser.py
```diff
@@ -229,3 +229,8 @@ def __init__(self):
             '-n', "--input-n-mer", default='25',
             help="Length of the peptide sequence to use when creating the FASTA from the pVACseq TSV.",
         )
+        self.parser.add_argument(
+            '--spacers', type=lambda s:[spacer for spacer in s.split(',')],
+            help="Comma-separated list of spacers to use for testing junction epitopes. Include an empty string (\"\") to test junctions without spacers.",
```
tmooney: This is really `include two double-quotes in a row`, isn't it?  An empty string would be nothing, like `BEFORE,,AFTER`.  It takes special care to send the quotes along in a shell (as longtime users of UR boolean expression are no doubt all-too-aware :smile:).

chrisamiller: Might it be better to just use a special string ("None"?) in place of an empty string, since there are so many ways to screw up the quoting? (quotes get eaten by job submissions, scripts, etc)

susannasiebert: I just pushed a commit to replace the `""` option with `None`. Unfortunately we can't use an empty string because then we wouldn't be able to submit just the no-spacer option.

lib/run_argument_parser.py
```diff
@@ -229,3 +229,8 @@ def __init__(self):
             '-n', "--input-n-mer", default='25',
             help="Length of the peptide sequence to use when creating the FASTA from the pVACseq TSV.",
         )
+        self.parser.add_argument(
+            '--spacers', type=lambda s:[spacer for spacer in s.split(',')],
+            help="Comma-separated list of spacers to use for testing junction epitopes. Include None to test junctions without spacers.",
+            default="None,HH,HHC,HHH,HHHD,HHHC,AAY,HHHH,HHAA,HHL,AAL"
```
chrisamiller: I don't think "None" should be in the default list

PR #270, opened 2019-01-24
Add locking around status-messages during multithreaded processing

### LINKED ISSUE #259: Parallel threads overwrite each other on stderr
* pvactools version:  1.2.0 - docker image

When running with 8 threads (via pvacvector), I see a few of these in the err log:

    Allele HLA-B*57:03 notRunning IEDB on Allele HLA-B*57:03 and Epitope Length 8 with Method NetMHCpan - Entries 1-2 - Completed

Presumably, this is due to threads both trying to write at the same time. It's low priority, but there oughta be a way to do some kind of blocking and prevent this from happening.

susannasiebert:
This is so different status messages from different threads don't clobber each other.

Closes #259

### DIFF
--- a/lib/pipeline.py
+++ b/lib/pipeline.py
@@ -26,8 +26,7 @@ def status_message(msg):
 
 def status_message_with_lock(msg, lock):
     lock.acquire()
-    print(msg)
-    sys.stdout.flush()
+    status_message(msg)
     lock.release()
 
 class Pipeline(metaclass=ABCMeta):


chrisamiller:
Nice! +1

lib/pipeline.py
```diff
@@ -18,11 +18,18 @@
...
+from threading import Lock
 
 def status_message(msg):
     print(msg)
     sys.stdout.flush()
 
+def status_message_with_lock(msg, lock):
+    lock.acquire()
+    print(msg)
+    sys.stdout.flush()
```
tmooney: This could call `status_message(msg)`.

PR #283, opened 2019-02-11
 Remove option to specify an additional input file list yaml when running pVACseq

### LINKED ISSUE #122: Remove the --additional-input-file-list option
https://github.com/griffithlab/pVACtools/pull/118 and https://github.com/griffithlab/pVAC-Seq/issues/219 added the functionality to have readcount and expression information directly parsed from the VCF annotations. This will ultimately result in the `--additional-input-file-list` yaml to become deprecated and removed. We should just go ahead and remove it from the api/client.

susannasiebert:
Since we now annotate and parse readcount and expression information in the VCF, we don't need to provide this option anymore.

Closes #122

### DIFF
--- a/lib/pipeline.py
+++ b/lib/pipeline.py
@@ -61,7 +61,7 @@ class Pipeline(metaclass=ABCMeta):
         self.keep_tmp_files              = kwargs.pop('keep_tmp_files', False)
         self.exclude_NAs                 = kwargs.pop('exclude_NAs', False)
         self.pass_only                   = kwargs.pop('pass_only', False)
-        self.normal_sample_name          = kwargs.pop('normal_sample_name', False)
+        self.normal_sample_name          = kwargs.pop('normal_sample_name', None)
         self.n_threads                   = kwargs.pop('n_threads', 1)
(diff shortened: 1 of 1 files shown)


lib/pipeline.py
```diff
@@ -165,21 +157,10 @@ def convert_vcf(self):
...
-            'normal_sample_name',
-        ]:
-            if getattr(self, attribute):
-                convert_params[attribute] = getattr(self, attribute)
-            else:
-                convert_params[attribute] = None
+        if self.normal_sample_name is not None:
+            convert_params['normal_sample_name'] = self.normal_sample_name
+        else:
+            convert_params['normal_sample_name'] = None
```
tmooney: Could this conditional be eliminated entirely? Looks like this is set to `self.normal_sample_name` either way.

susannasiebert: This is actually wrong the way I wrote this. The else clause isn't supposed to be in there. I removed it.

susannasiebert:
> Are all the files under the `tmp*`-named directories like `tmpdlzgqv9_` meant to be included in this PR?

Yeah, that was an accident. I rebased to remove them.

PR #285, opened 2019-02-14
Suppress Tensorflow warning

susannasiebert:
This will remove the annoying "Using TensorFlow backend." status message whenever you execute any pvactools command. It will also clean up our documentation (once we have a new pVACtools release with this change and rebuild the docs).

### DIFF
--- a/lib/prediction_class.py
+++ b/lib/prediction_class.py
@@ -3,8 +3,11 @@ import os
 import csv
 import sys
 import inspect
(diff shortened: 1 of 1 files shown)


susannasiebert:
Seriously. I tried everything else suggested like adjusting various warning level parameters and this was the only thing that worked 🤷‍♀️

chrisamiller:
+1

PR #293, opened 2019-02-20
Fix indentation of code to have it execute at correct level of nested for-loop

susannasiebert:
This bug would lead to errors such as `CRITICAL:pymp:An exception occured in thread 1: (<class 'FileNotFoundError'>, [Errno 2] No such file or directory: '/cromwell-executions/pvacseq.cwl/eb9445e6-d282-4ec6-8406-5ceaacde3367/call-pvacseq.cwl/execution/MHC_Class_I/tmp/TUMOR.HLA-C*06:02.10.parsed.tsv_1601-1800.tmp' -> '/cromwell-executions/pvacseq.cwl/eb9445e6-d282-4ec6-8406-5ceaacde3367/call-pvacseq.cwl/execution/MHC_Class_I/tmp/TUMOR.HLA-C*06:02.10.parsed.tsv_1601-1800')` when running with more than one thread. It would also lead to erroneous log messages such as `Parsed Output File [..] already exists. Skipping.`

### DIFF
--- a/lib/pipeline.py
+++ b/lib/pipeline.py
@@ -358,6 +358,7 @@ class Pipeline(metaclass=ABCMeta):
                                 if os.path.getsize(split_fasta_file_path) == 0:
                                     status_message_with_lock("Fasta file is empty. Skipping", lock)
                                     continue
+                                #begin of per-algorithm processing
                                 with pymp.Parallel(iteration_info['algorithm']['threads']) as p4:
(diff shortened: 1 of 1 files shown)


chrisamiller:
+1

susannasiebert:
@tmooney eh not your fault. I didn't expect it to have this side effect since it was still executing after the for-loop but just inside of the multithreading context (instead of outside). Glad it got figured it out though because that error was driving me bananas.

Issue #291, opened 2019-02-19
add command-line option to supply IP address to api and viz


jmcmichael:
Amazon and maybe other cloud providers have a public/private IP system, and only the private IP is discovered by the startup scripts. Since the public IP will need to be used for public access, we'll need to provide a method for users to manually provide this IP. I'm thinking --ip-address [address] would work.

malachig:
This would be a new option that you would specify when you start the `pvacviz` process?

susannasiebert:
@malachig that's my understanding.

PR #301, opened 2019-02-22
adds command-line arg for ip_address to pvacapi/viz

jmcmichael:
In order to account for server environments where the public IP is not easily available to running python processes (e.g. AWS servers), this PR adds a ip_address argument to both pvacapi and viz startup scripts.

If no ip_address is supplied, the scripts will bind to the address returned from the getIPAddress() function as before.

### DIFF
--- a/utils/pvacapi/app.py
+++ b/utils/pvacapi/app.py
@@ -14,8 +14,8 @@ from utils.pvacapi.controllers.utils import getIpAddress
 #FIXME: sanitize sample name
 def main():
     parser = argparse.ArgumentParser(description='pVACapi provides a REST API to pVACtools')
-    parser.add_argument('--ip_address', help='IP address for the HTTP server to bind. If not provided, the default socket address will be used.')
-    parser.add_argument('--proxy_ip_address', help='IP address of proxy server or public IP address. If provided, server will send X-Forward headers required for Bokeh to properly work through a proxy server or with AWS private/public IP addresses.')
+    parser.add_argument('--ip-address', help='IP address for the HTTP server to bind. If not provided, the default socket address will be used.')
+    parser.add_argument('--proxy-ip-address', help='IP address of proxy server or public IP address. If provided, server will send X-Forward headers required for Bokeh to properly work through a proxy server or with AWS private/public IP addresses.')
     parser.add_argument('--debug', default=False, action='store_true', help='Start sever in debug mode.')
     args = parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


susannasiebert:
Closes #291

utils/pvacapi/app.py
```diff
@@ -3,17 +3,18 @@
...
+import argparse
 
-from webbrowser import open_new_tab
 from flask_cors import CORS
 from utils.pvacapi.controllers.utils import initialize
 from utils.pvacapi.controllers.utils import getIpAddress
+
 #FIXME: sanitize sample name
 def main():
+    parser = argparse.ArgumentParser(description='parse pvacapi arguments')
```
susannasiebert: The description actually gets output as part of the usage information. So this should be something more descriptive about pVACapi itself.

utils/pvacviz/app.py
```diff
@@ -35,11 +35,20 @@ def do_GET(self):
 
 def main():
     os.chdir(CLIENTDIR)
+    parser = argparse.ArgumentParser(description='parse pvacviz arguments')
```
susannasiebert: Same here about description.

utils/pvacapi/app.py
```diff
@@ -27,13 +28,18 @@ def main():
     class IntConverter(BaseIntConverter):
         regex = r'-?\d+'
 
+    # determine IP address and setup CORS
+    IPAddr = None
+    if args.ip_address is None:
+        IPAddr = getIpAddress()
```
susannasiebert: pVACapi doesn't do a very good job following PEP8 conventions but variable names should be under_score, not camelCase.

utils/pvacviz/app.py
```diff
@@ -35,11 +35,20 @@ def do_GET(self):
...
+    parser.add_argument('--ip_address', help='IP address for the viz HTTP server to bind')
+    args = parser.parse_args()
+
+    IPAddr = None
```
susannasiebert: Same here about variable names.

utils/pvacviz/app.py
```diff
@@ -35,11 +35,20 @@ def do_GET(self):
 
 def main():
     os.chdir(CLIENTDIR)
+    parser = argparse.ArgumentParser(description='parse pvacviz arguments')
+    parser.add_argument('--ip_address', help='IP address for the viz HTTP server to bind')
```
susannasiebert: For other commands we've been using dash-separated naming of optional arguments. So this should be changed to `--ip-address`.

susannasiebert:
I had made some code review suggestions here that still need to be addressed. Also, the branch now also has merge conflicts that will need to be resolved.

PR #304, opened 2019-02-25
Test one spacer at a time when running pVACvector

### LINKED ISSUE #300: First pass at speeding up pVACvector binding predictions
Instead of testing each peptide pair with all possible spacer combinations, let's make binding predictions with one spacer at a time. * If pVACvector finds a valid path with those predictions, it will exit and return that path. If it doesn't it will make binding predictions with the next spacer in the list. The binding predictions will be appended to the results from the previous run(s) and pVACvector will try to find a valid path over that whole set. Repeat from *.  

The input order of the spacers is the order that spacers will be tested. 

To diversity the list of spacers to test, we will update the default spacer list to the following: `None,AAY,HHHH,GGS,GPGPG,HHAA,AAL,HH,HHC,HHH,HHHD,HHL,HHHC`

susannasiebert:
Closes #300

### DIFF
--- a/docs/pvacvector.rst
+++ b/docs/pvacvector.rst
@@ -10,10 +10,21 @@ DNA vector based personalized cancer vaccines. It takes as input either a pVACse
 tsv file or a fasta file containing peptide sequences and returns a peptide ordering that
 minimizes the effects of junctional epitopes (that may create novel peptides)
 between the sequences. It does this by using the core pVACseq services to
-predict the binding scores for each junctional peptide. It also tests
-junctions with spacer amino acid sequences that may help to eliminate junctional epitopes.
-These spacer amino acid sequences can be "HH", "HHC", "HHH", "HHHD", "HHHC",
-"AAY", "HHHH", "HHAA", "HHL" or "AAL. The final vaccine ordering is
+predict the binding scores for each junctional peptide separated by a spacer amino acid
+sequence that may help to eliminate junctional epitopes. The list of spacers
+to be tested is specified using the ``--spacers`` parameter. Peptide combinations without a
+spacer can be tested by including ``None`` in the list of spacers.
+
+Peptide junctions are tested with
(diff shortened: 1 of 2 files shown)


chrisamiller:
Seems like a nice extension of what was already there, and I can't think of any edge cases that would choke it at the moment. +1 

Have you had a chance to try it on anything other than your small test cases? If not, maybe @jhundal and Mike M would be willing to run a beta version on some recent cases as a sanity check?

susannasiebert:
I only ran it on my test file but with different binding affinity cutoffs to test that spacers get added as expected and it seems to be working but yes, it would be good to test this on some real data as well.

susannasiebert:
TODO: Update documentation

docs/pvacvector.rst
```diff
@@ -10,10 +10,21 @@ DNA vector based personalized cancer vaccines. It takes as input either a pVACse
 tsv file or a fasta file containing peptide sequences and returns a peptide ordering that
 minimizes the effects of junctional epitopes (that may create novel peptides)
 between the sequences. It does this by using the core pVACseq services to
-predict the binding scores for each junctional peptide. It also tests
-junctions with spacer amino acid sequences that may help to eliminate junctional epitopes.
-These spacer amino acid sequences can be "HH", "HHC", "HHH", "HHHD", "HHHC",
-"AAY", "HHHH", "HHAA", "HHL" or "AAL. The final vaccine ordering is
+predict the binding scores for each junctional peptide separated by a spacer amino acid
+sequences that may help to eliminate junctional epitopes. The list of spacers
```
tmooney: I'd go with either `by a spacer amino acid sequence` or `by spacer amino acid sequences`.
```suggestion
sequence that may help to eliminate junctional epitopes. The list of spacers
```

susannasiebert: I went with the first suggestion just because there is only one spacer used at a time.

PR #313, opened 2019-03-01
Parse binding prediction files after making all binding predictions

susannasiebert:
This is not as well-done as I would like it to be but this should solve the problem we are seeing with truncated output files. This PR splits the parsing of prediction tsv files from making the prediction itself. Making binding predictions is still multithreaded, while the parsing steps are once again in serial. This should also help with memory issues since parsing is what takes up a lot of memory because we are reading the whole prediction file into memory. Parsing in parallel was, thus, using a lot of memory but going back to parsing serially should alleviate some of these problems.

I'm basing this off of staging for now since Mike M needs this as well as the pVACvector spacer improvements but we need to make a hotfix 1.3.2 release for this as well.

I'm currently running Connor's VCF with this fix locally but it will be a while before it finishes since his VCF is so large.

### DIFF
--- a/lib/pipeline.py
+++ b/lib/pipeline.py
@@ -307,7 +307,7 @@ class Pipeline(metaclass=ABCMeta):
                 return iteration_info
         return iteration_info
 
-    def call_iedb_and_parse_outputs(self, chunks):
+    def call_iedb(self, chunks):
         pymp.config.nested = True
         alleles = self.alleles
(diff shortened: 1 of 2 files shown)


chrisamiller:
Seems sensible - just pulling that code out of the multi-threaded chunk is fine for now (and maybe permanently, given how fast it is compared to the binding predictions).  +1

PR #317, opened 2019-03-06
Replace status_message_with_lock with pymp's print

susannasiebert:
We encountered some zombie processes when using multi-threading. As far as @tmooney was able to figure out, there was one child process that disappeared while it held the print lock so the other child processes were waiting on it, halting the whole process. By switching to the pymp print, I hope that this issue can be circumvented.


chrisamiller:
+1

PR #320, opened 2019-03-12
Improve status messages when no valid Path is found

susannasiebert:
Previously we would only write the status message "No valid path found." without any indication as to why. Now, the status message will output problematic peptides which can be used to modify the input fasta.

### DIFF
--- a/tools/pvacvector/run.py
+++ b/tools/pvacvector/run.py
@@ -187,25 +187,22 @@ def check_graph_valid(Paths, seq_dict):
         errors.append("No valid junctions found for peptides: {}".format(set(seq_dict.keys()) - set(Paths.nodes())))
 
     nodes_without_outgoing_edges = []
+    nodes_without_incoming_edges = []
+    nodes_without_any_edges = []
     for node in Paths.nodes():
         if len(Paths.out_edges(node)) == 0:
             nodes_without_outgoing_edges.append(node)
+        if len(Paths.in_edges(node)) == 0:
+            nodes_without_incoming_edges.append(node)
+        if len(Paths.in_edges(node)) == 0 and len(Paths.out_edges(node)) == 0:
+            nodes_without_any_edges.append(node)
+
     if len(nodes_without_outgoing_edges) > 1:
         graph_valid = False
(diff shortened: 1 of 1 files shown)


tools/pvacvector/run.py
```diff
@@ -176,36 +176,35 @@ def create_graph(iedb_results, seq_tuples, spacers):
...
-    n_nodes_without_any_edges = 0
+    nodes_without_any_edges = []
     for node in Paths.nodes():
         if len(Paths.in_edges(node)) == 0 and len(Paths.out_edges(node)) == 0:
-            n_nodes_without_any_edges += 1
-    if n_nodes_without_any_edges > 0:
-        return False
+            nodes_without_any_edges.append(node)
+    if len(nodes_without_any_edges) > 0:
+        return (False, "No valid junctions found for peptides: {}".format(nodes_without_any_edges))
```
tmooney: Is it possible to reach this `return`? Seems like the condition here would've necessarily triggered both of the previous errors.

Would it be beneficial to the user to go through the nodes populating both  `nodes_without_outgoing_edges` and `nodes_without_incoming_edges`, then checking and reporting on both together?

tools/pvacvector/run.py
```diff
@@ -176,36 +176,35 @@ def create_graph(iedb_results, seq_tuples, spacers):
...
 
-    return True
+    return (True, "")
```
tmooney: Would it make sense to use `None` for the condition where there's no error?

tools/pvacvector/run.py
susannasiebert: In the two previous checks you could've ended up with one node without incoming edges and one node without outgoing edges. If those two nodes are the same node, this condition would trigger. However, I don't _think_ we would even add this node to the graph to begin with so this would get caught by the very first check. I'll probably leave it in just to be sure.

Yeah, it would be best to report on the full set.

tmooney: Ah! What I was missing here was the `0` in this condition versus the `1` in the others... so single node is still caught here.  Makes sense, thanks! :smile:

chrisamiller:
+1

susannasiebert:
@tmooney I implemented the suggestions you made.

PR #321, opened 2019-03-12
Repeat pVACvector processing with shortened sequences if first try doesn't find valid path

susannasiebert:
We've been running into situations where none of the spacers yield a combination with valid junctions. Removing amino acids from the beginning and/or end of problematic peptides. This will yield new/different junction epitopes that might not be as well-binding.

Right now this feature is agnostic to where in the peptide sequence the vaccine epitope is located so it has the potential to remove amino acid from the actual vaccine epitope, which is bad. This should be improved upon in future iterations.

TODO: make `max_tries` configurable (parameter naming suggestions welcome).

### DIFF
--- a/lib/run_argument_parser.py
+++ b/lib/run_argument_parser.py
@@ -231,3 +231,8 @@ class PvacvectorRunArgumentParser(RunArgumentParser):
             help="Comma-separated list of spacers to use for testing junction epitopes. Include None to test junctions without spacers. Peptide combinations will be tested with each spacer in the order specified.",
             default="None,AAY,HHHH,GGS,GPGPG,HHAA,AAL,HH,HHC,HHH,HHHD,HHL,HHHC"
         )
+        self.parser.add_argument(
+            '--max-clip-length', type=int,
+            help="Number of amino acids to permit clipping from the start and/or end of peptides in order to test novel junction epitopes when the first pass on the full peptide fails.",
+            default=3,
+        )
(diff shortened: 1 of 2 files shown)


lib/run_argument_parser.py
```diff
@@ -231,3 +231,8 @@ def __init__(self):
             help="Comma-separated list of spacers to use for testing junction epitopes. Include None to test junctions without spacers. Peptide combinations will be tested with each spacer in the order specified.",
             default="None,AAY,HHHH,GGS,GPGPG,HHAA,AAL,HH,HHC,HHH,HHHD,HHL,HHHC"
         )
+        self.parser.add_argument(
+            '--max-clip-length', type=int,
+            help="Number of amino acids to permit clipping from the start and/or end of peptides in order to test novel junction epitopes when the first pass on the full peptide fails.",
+            default=3,
+        )
```
chrisamiller: My initial inclination was to limit the final peptide size, rather than limit the clipping length, but I think I like this better upon review.  It's intuitive and simplifies a lot of the code.

susannasiebert: I do have a commit locally to would use a minimum_peptide_length instead of the max_clip_length. Do you want me to push it so you can see how it would be different? Maybe in a separate PR?

susannasiebert:
@chrisamiller we had at one point discussed to only attempt clipping if more than 75% of peptides have good junctions. Right now it clips no matter what. Should I still implement this check?

lib/run_argument_parser.py
chrisamiller: yeah, sure - I think either could work, and the min_peptide_length might be more intuititive ot people

chrisamiller:
As long as it's doing all the spacer checks before trying clipping (and it is, if I'm reading this right) then sure, I think it's fine

PR #323, opened 2019-03-13
Change --max-clip-length to --minimum-peptide-length

susannasiebert:
@chrisamiller here is the alternative implementation using a minimum peptide length instead of max clip length.

### DIFF
--- a/tools/pvacvector/run.py
+++ b/tools/pvacvector/run.py
@@ -286,7 +286,9 @@ def shorten_problematic_peptides(input_file, problematic_start, problematic_end,
             records.append(record)
         else:
             if record.id in problematic_start and record.id in problematic_end:
-                new_seq = str(record.seq)[1:-1]
+                new_seq = str(record.seq)[1:]
+                if len(new_seq) > min_peptide_length:
+                    new_seq = new_seq[:-1]
             elif record.id in problematic_start:
                 new_seq = str(record.seq)[1:]
             elif record.id in problematic_end:


tools/pvacvector/run.py
```diff
@@ -278,23 +278,33 @@ def find_optimal_path(Paths, distance_matrix, seq_dict, seq_keys, base_output_di
...
+        if record.id not in problematic_start and record.id not in problematic_end:
+            records.append(record)
         else:
-            record_new = record
-        records.append(record_new)
-    os.makedirs(output_dir, exist_ok=True)
-    new_input_file = os.path.join(output_dir, "vector_input.fa")
-    SeqIO.write(records, new_input_file, "fasta")
-    return new_input_file
+            if record.id in problematic_start and record.id in problematic_end:
```
chrisamiller: Let's say that we have a peptide that's 33aa and a max-length of 32aa.  If both the start and end are labelled problematic, we will never do any trimming of either end (because it will try to trim both at once, and drop below the min length).  It would be nice to alter this such that it trims the front first, then checks length, then trims the end, then checks length again (so at least one end would get trimmed in the scenario I laid out).

susannasiebert:
I just pushed two commits to address your comments.

chrisamiller:
looks good! +1

susannasiebert:
@chrisamiller do you prefer the `--minimum-peptide-length` over `--max-clip-length`?

chrisamiller:
I keep waffling on this one - I can see the appeal of both, but I think at the end of the day, `--max-clip-length` feels a little more intuitive (despite what I said originally - sorry...).  I could definitely still be persuaded either way if someone has stronger feelings

susannasiebert:
The current implementation of `--max-clip-length` is agnostic as to which peptides have been clipped. Right now it just repeats the clipping process three times. Since the list of "bad" peptides can change in between successive clippings, we can end up with peptides that could be clipped more to meet the max clip length. So if we decide that we want to go with max clip length over min peptide length, that's something I'd want to improve upon.

chrisamiller:
@jhundal - do you have a preference here?

jhundal:
My initial inclination was to go with `--max-clip-length`, but now based on Susanna's comment, it seems like  we can end up with peptides that could be clipped more over successive runs to determine the optimal solution. We certainly do not want to 'over' clip incase we end up clipping the actual neoepitope. So I'd suggest using `--minimum-peptide-length`, though I'm conflicted about this one as well. 

The ideal solution (even if it takes longer to implement) would be to check for the exact neoepitope peptide sequence when a TSV input is provided, and always make sure we do not end up clipping that. This would most certainly be helpful for Class II, since those are generally longer peptides (15mers+).

chrisamiller:
oh wait - a thought:
on peptides right at the beginning of a protein, there may not be spacing around it
on one side (or the other, if it's the end).  So we may need a pair of params:  `--max-clip-left` and `--max-clip-right`

PR #339, opened 2019-03-26
Cast start and end positions to integers

susannasiebert:
When manipulating the start and end positions we might end up with
decimal representation of whole numbers which will cause an error when
using them to extract the peptide. Casting them to integer first will
eliminate this problem.

This also fixes and issue where the start position was < 0, which would result in incorrect/empty peptides getting extracted.

TODO:  Add a test!


chrisamiller:
+1

PR #347, opened 2019-04-03
Don't parallelize MHCflurry or MHCnuggets

susannasiebert:
MHCflurry and MHCnuggets (more precisely tensorflow) seem to be incompatible with multithreading, at least in the way that we are using them currently. https://github.com/tensorflow/tensorflow/issues/8220 has some suggestions to make it work but in the meantime removing MHCnuggets and MHCflurry from multithreading is a stop-gap measure that should avoid stalled runs.


chrisamiller:
+1 This seems like a sensible workaround!

susannasiebert:
@tmooney I think this will be the long-term solution. But for now I want to get this out so that we have a working version.

PR #350, opened 2019-04-05
Write log files of junction scores for each attempted iteration

### LINKED ISSUE #346: Output a file with all junction scores each time we create a graph in pVACvector
This will make it much easier to debug which junctions/peptides are bad.

susannasiebert:
Closes #346


tools/pvacvector/run.py
```diff
@@ -112,8 +112,62 @@ def run_pipelines(input_file, base_output_dir, args, spacer):
...
+                'right_peptide': right_peptide,
+                'epitope': row['MT Epitope Seq'],
+                'allele': row['HLA Allele'],
+            }
+            if args.top_score_metric == 'lowest':
+                new_row['junction_score'] = float(row['Best MT Score'])
+                new_row['method'] = row['Best MT Score Method']
+            elif args.top_score_metric == 'median':
+                new_row['junction_score'] = float(row['Median MT Score'])
+                new_row['method'] = 'NA'
```
chrisamiller: instead of NA, would it be better to fill this with "median"?

susannasiebert: Sure. Do you think they columns are ok the way they are or is there any info you think would be good to also display?

chrisamiller: I think this is probably the right level of granularity. We can always parse the tmp epitope files if we really need additional info

PR #352, opened 2019-04-08
Improve retry process for calling the IEDB API

susannasiebert:
The IEDB API returns a 403 - Forbidden when too many requests are made at the same time. This is intermittent so this PR will make it so that we retry on this return code.

This PR also adds some randomness to the wait between retries in order to not have retries hit at the same time during multiprocessing. random.seed(uuid.uuid4().int) is used because on certain systems the timestamp seeds the random number generator which will lead to the same "random" numbers getting generated in parallel processes. uuid is truly unique so the seed will be unique for each parallel process and the wait will be different between parallel processes.

### DIFF
--- a/lib/prediction_class.py
+++ b/lib/prediction_class.py
@@ -20,6 +20,8 @@ from subprocess import run, PIPE
 import tempfile
 from collections import defaultdict
 from Bio import SeqIO
+import random
+import uuid
 
 class IEDB(metaclass=ABCMeta):
     @classmethod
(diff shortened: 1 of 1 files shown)


lib/prediction_class.py
```diff
@@ -79,11 +81,12 @@ def predict(self, input_file, allele, epitope_length, iedb_executable_path, iedb
 
             response = requests.post(self.url, data=data)
             retries = 0
-            while response.status_code == 500 and retries < iedb_retries:
-                time.sleep(60 * retries)
-                response = requests.post(self.url, data=data)
-                print("IEDB: Retry %s of %s" % (retries, iedb_retries))
+            while (response.status_code == 500 or response.status_code == 403) and retries < iedb_retries:
+                random.seed(uuid.uuid4().int)
```
tmooney: Shouldn't need to re-seed on every iteration of the loop, but it's also mostly harmless.

PR #365, opened 2019-04-15
Sort report rows before running the top score filter

### LINKED ISSUE #362: Ensure ordering of epitopes before running the top score filter
Right now we order the output for the all_epitopes file by Gene - Mutation. However, subsequent filters use pandas query and it's unclear whether that preserves the order of the input data. Either way we can't count on an ordered file being provided to the standalone top score filter so it would be prudent to sort the entries by Gene - Mutation again before running the top score filter.

susannasiebert:
Closes #362

### DIFF
--- a/lib/top_score_filter.py
+++ b/lib/top_score_filter.py
@@ -1,5 +1,6 @@
 import csv
 import argparse
+import lib.sort
 
 class TopScoreFilter:
     def __init__(self, input_file, output_file, top_score_metric):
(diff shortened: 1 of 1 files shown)


susannasiebert:
@tmooney while `top_score_metric` is a argument to this function, the sorting is actually on five columns. `top_score_metric` is simply a switch to pick Median vs Best MT Score as one of the five sort columns.

PR #376, opened 2019-04-19
Add documentation for common pVACseq errors

### LINKED ISSUE #374: Add documentation on each of the errors pVACseq might throw
This will be a space to add more details for each error and make suggestions how to fix it (if possible).

susannasiebert:
Closes #374

### DIFF
--- a/docs/pvacseq/errors.rst
+++ b/docs/pvacseq/errors.rst
@@ -141,8 +141,9 @@ mismatch cannot be resolved by pVACseq, which is why this error is fatal.
 
 Here are a few things that might resolve this error:
 
-- If a VEP build 38 cache was accidentially used with a build 37 VCF (or vice
-  versa), the correct cache needs to be downloaded and used during VEP annotation.
+- Checking that the build of the VEP cache matches the alignment build and
+  downloading the correct cache if there is a build mismatch (such as a build
+  38 cache with a build 37 VCF, or vice versa)
 - Using the ``--assembly`` parameter during VEP annotation with the
   correct build version to match your VCF
 - Using the ``fasta`` parameter during VEP annotation with the reference used


docs/pvacseq/errors.rst
```diff
@@ -0,0 +1,169 @@
+Common Errors
+-------------
+
+Input VCF Sample Information
+____________________________
+
+**VCF contains more than one sample but sample_name is not set.**
+
+pVACseq supports running with a multi-sample VCF as input. However, it still
```
tmooney: I would avoid using `still` here unless the `sample_name` property is required when the VCF has only one sample, and we are clarifying that it is required in this case, too.
```suggestion
pVACseq supports running with a multi-sample VCF as input. However, in this case it
```

docs/pvacseq/errors.rst
```diff
@@ -0,0 +1,169 @@
...
+requires the user to pick the sample to analyze, as only variants that are
+called in a sample will be processed.
```
tmooney: If we're saying it's limited to the chosen sample, I'd make that more clear.  Variants called only in other samples were called in `a` sample...just not the one we care about.
```suggestion
called in the specified sample will be processed.
```

docs/pvacseq/errors.rst
```diff
@@ -0,0 +1,169 @@
...
+
+All columns after ``FORMAT`` are sample identifiers that can be used as the
+``--normal-sample-name`` parameter when running pVACseq, depending on which
+sample is the normal sample in the VCF. Change the ``--normal-sample-name`` parameter of your ``pvacseq
+run`` command to match the appropriate sample identifier.
+
+**VCF doesn't contain any sample genotype information.**
+
+pVACseq uses the sample genotype to identified which variants were called.
+Therefore, while a VCF with a ``FORMAT`` and sample column(s) is valid, it cannot be used
```
tmooney: We're saying that we need a `FORMAT`, right?
```suggestion
Therefore, while a VCF without a ``FORMAT`` and sample column(s) is valid, it cannot be used
```

docs/pvacseq/errors.rst
```diff
@@ -0,0 +1,169 @@
...
+bgzipped and tabix indexed. See :ref:`the Input File
+Preparation section <bgzip_phased_vcf>` of the documentation for instructions on how to do so.
+
+Input VCF VEP Annotation
+________________________
+
+**Input VCF does not contain a CSQ header. Please annotate the VCF with VEP before running it.**
+
+pVACseq requires the input VCF to be annotated by VEP. The provided input VCF
+doesn't contain a ``CSQ`` ``INFO`` header which indicates that it has not been
```
tmooney: This sentence is a little confusing in that the `CSQ` header indicates that the VCF *has* been annotated, but the sentence is talking about how a lack of it indicates that is has not.  It's easy to read this as saying the opposite of what it does.

I'd at least add a comma:
```suggestion
doesn't contain a ``CSQ`` ``INFO`` header, which indicates that it has not been
```
Maybe more separation would help:
```suggestion
doesn't contain a ``CSQ`` ``INFO`` header. This indicates that it has not been
```

docs/pvacseq/errors.rst
```diff
@@ -0,0 +1,169 @@
...
+annotated. :ref:`The Input File Preparation section <vep>` of the
+documentation provides instructions on how to annotate your VCF with VEP.
+
+**VCF doesn't contain VEP DownstreamProtein annotations. Please re-annotate the VCF with VEP and the Wildtype and Downstream plugins.**
+
+Although the input VCF was annotated with VEP, it is missing the required
+annotations provided by the VEP Downstream plugin. The input VCF will need to
+be reannotated using all of the required arguments as outline in the :ref:`Input
```
tmooney:
```suggestion
be reannotated using all of the required arguments as outlined in the :ref:`Input
```

docs/pvacseq/errors.rst
```diff
@@ -0,0 +1,169 @@
...
+File Preparation section <vep>` of the documentation.
+
+**VCF doesn't contain VEP WildtypeProtein annotations. Please re-annotate the VCF with VEP and the Wildtype and Downstream plugins.**
+
+Although the input VCF was annotated with VEP, it is missing the required
+annotations provided by the VEP Wildtype plugin. The input VCF will need to
+be reannotated using all of the required arguments as outline in the :ref:`Input
```
tmooney:
```suggestion
be reannotated using all of the required arguments as outlined in the :ref:`Input
```

docs/pvacseq/errors.rst
```diff
@@ -0,0 +1,169 @@
...
+File Preparation section <vep>` of the documentation.
+
+**Proximal Variants VCF does not contain a CSQ header. Please annotate the VCF with VEP before running it.**
+
+When running pVACseq with the ``--proximal-variants-vcf`` argument, that
+proximal variants VCF needs to be annotated by VEP. The provided proximal
+variants VCF
+doesn't contain a ``CSQ`` ``INFO`` header which indicates that it has not been
```
tmooney: A couple possible choices just like the other instance of this.
```suggestion
doesn't contain a ``CSQ`` ``INFO`` header, which indicates that it has not been
```

```suggestion
doesn't contain a ``CSQ`` ``INFO`` header. This indicates that it has not been
```

docs/pvacseq/errors.rst
```diff
@@ -0,0 +1,169 @@
...
+prediction is based on the Ensembl transcript nucleotide sequence.
+
+This points to a fundamental disagreement between the reference that was
+used during alignment and variant calling and the Ensembl reference. This
+mismatch cannot be resolved by pVACseq, which is why this error is fatal.
+
+Here are a few things that might resolve this error:
+
+- If a VEP build 38 cache was accidentially used with a build 37 VCF (or vice
+  versa), the correct cache needs to be downloaded and used during VEP annotation.
```
tmooney: This list item breaks parallelism with the other list items.  It would need to be phrased so as to say something like `If a VEP build 38 cache was accidentially used with a build 37 VCF (or vice
  versa), downloading the correct cache and using it during VEP annotation`.

That particular phrasing is a bit cumbersome.  Maybe something like:
`checking that the build of the VEP cache matches the alignment build and downloading the correct cache if there is a build mismatch (such as a build 38 cache with a build 37 VCF, or vice versa)`

Alternatively, all the other list items could have their verb tenses changed to match this one!

PR #383, opened 2019-04-30
New post-processing step to calculate manufacturability metrics

### LINKED ISSUE #280: Create helper tool for selection of long peptide sequences
It would be helpful to have a tool that helped end users optimize selection of long peptide sequences.  Specifically for those who are making long peptide vaccines.

e.g. Imagine you have a 9-mer that you like.  You want to choose a 22-25mer long peptide that contains this 9-mer.  We have some wiggle room for both the length and sub-peptide position we chose when submitting to the manufacturer.  Certain characteristics need to be avoided and this can be done by adjusting either the length or register.  For example, a long peptide should not start or end with a Proline and Glutamine.  Also in general you want a longer peptide because it increases solubility on average.  However, this must be balanced against an increased *synthesis* failure rate for longer peptides.  Right now people are looking at these sequences and picking them manually to account for these factors.  In this step we may incorporate a more sophisticated calculation of *manufacturability* (e.g. like in vaxrank (https://github.com/openvax/vaxrank/issues/2)).

Obviously we would need to refine the details more here.  This is just to get the discussion rolling.

susannasiebert:
Closes #280


chrisamiller:
+1 on the code

Do we have any guidance on how to actually use these metrics? Is there a obvious way to create an overall "manufacturability score" that can be used for prioritization?  

Presumably vaxrank has some docs of how to interpret these somewhere - can we update the documentation to point to some of that?

susannasiebert:
Not yet. We were hoping to correlate some of these metrics to the the manufacturability information we got from the actual vendors (?) in the hopes of coming up with something. Vaxrank uses these metrics to filter but I haven't found any documentation on their exact workflow.

susannasiebert:
https://www.biorxiv.org/content/10.1101/142919v2.full has a section on Manufacturability.

CodyRamirez:
When we ordered long peptides from our vendor we followed these same basic rules which is simply to stay away from certain amino acids and to consider hydropathy. You can see a bit more detail notes about how vaxrank uses it here: https://github.com/openvax/vaxrank/blob/69100b7265177633bd6c648d58d0272c5362b326/vaxrank/vaccine_peptide.py#L80-L105

lib/calculate_manufacturability.py
```diff
@@ -0,0 +1,51 @@
...
+                peptide = PvacpeptideVaccinePeptide(line['MT Epitope Seq'])
+                metrics = peptide.manufacturability_scores
+                line['cterm_7mer_gravy_score'] = metrics.cterm_7mer_gravy_score
+                line['max_7mer_gravy_score'] = metrics.max_7mer_gravy_score
+                line['difficult_n_terminal_residue'] = metrics.difficult_n_terminal_residue
+                line['c_terminal_cysteine'] = metrics.c_terminal_cysteine
+                line['c_terminal_proline'] = metrics.c_terminal_proline
+                line['cysteine_count'] = metrics.cysteine_count
+                line['n_terminal_asparagine'] = metrics.n_terminal_asparagine
+                line['asparagine_proline_bond_count'] = metrics.asparagine_proline_bond_count
```
tmooney: Part of me thinks it'd be nice to do something like this:
```python
    for metric in self.manufacturability_headers():
        line[metric] = getattr(metrics, metric)
```
This would make it harder to have the headers diverge from the gathered metrics.  On the other hand, the way it's already listed out here maybe makes it easier to follow.

PR #405, opened 2019-05-29
 Convert the pVACseq score to a rank

### LINKED ISSUE #392: Convert pVACseq to a rank and output that
We don't want users to read too much into the score so converting it to a rank would be better.

susannasiebert:
Closes #392

### DIFF
--- a/docs/pvacseq/features.rst
+++ b/docs/pvacseq/features.rst
@@ -99,7 +99,7 @@ for each variant. As with all previous filters, this filter can also be run
 
 **Ranking of candidate neoepitopes**
 
-Filtered neoepitopes are :ref:`ranked <score>` based on the binding affinity,
+Filtered neoepitopes are :ref:`ranked <rank>` based on the binding affinity,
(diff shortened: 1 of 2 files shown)


chrisamiller:
+1

docs/pvacseq/features.rst
```diff
@@ -97,9 +97,9 @@ As a last filtering step, pVACseq applies the top score filter to only keep the
 for each variant. As with all previous filters, this filter can also be run
 :ref:`standalone <filter_commands>`.
 
-**Scoring of candidate neoepitopes**
+**Ranking of candidate neoepitopes**
 
-Filtered neoepitopes are :ref:`scored and ranked <score>` based on the binding affinity,
+Filtered neoepitopes are :ref:`ranked <score>` based on the binding affinity,
```
tmooney: Should we update this reference (`<score>`) as well?  (Changing it breaks old URLs, I guess, but keeping it keeps the word `score` around in them instead.)

PR #406, opened 2019-05-30
Resolved bugs relating to concurrent database accesses

atwollam:
There would often times be errors when (a) user(s) would try and visualize multiple files at the same time.  This change makes such concurrent use execute without issue.

### DIFF
--- a/utils/pvacapi/controllers/database.py
+++ b/utils/pvacapi/controllers/database.py
@@ -87,6 +87,120 @@ def column_mapping(row, mapping, schema):
             output[col] = None
     return (mapping, output, changes)
 
+def create_table(parentID, fileID, data, tablekey, db):
+    # Open a reader to cache the file in the database
+    if parentID != -1:
+        process = fetch_process(parentID, data, current_app.config['storage']['children'])
+        if not process[0]:
+            return (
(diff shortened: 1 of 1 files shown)


utils/pvacapi/controllers/database.py
```diff
@@ -323,7 +330,7 @@ def fileschema(parentID, fileID):
     )
 
     # check if the table exists:
-    db = current_app.config['storage']['db']
+    db = psql.open("localhost/pvacseq")
```
susannasiebert: Is it correct to use `localhost` here? We've been making some changes so that people can run pvacapi with any arbitrary ip address (`--ip-address` argument).

atwollam: It should be correct to use `localhost` here.  This line of code creates a connection to the locally run postgres database server, which is run at `localhost/pvacseq`.  As of the current version of staging, this is how the database is setup to run in `utils.py`.

susannasiebert: Ah, you're correct. Thanks for the clarification. I still would like to see the refactor to move some of the code into their own subroutines, if possible but otherwise +1

PR #408, opened 2019-06-11
Database File Read Speed Increase

atwollam:
This PR changes the way in which visualizable files are read into the database.  Primarily, these changes cause the speed of the read to be increased.

Additionally, It was noted that the "protein_position" field in visualizable files was improperly visualized in older versions.  This field is sometimes expressed in the form of a range (i.e. "283-284") rather than just an integer, which makes point-based visualization problematic.  As such, in decision with @jmcmichael , "protein_position" has been temporarily removed from visualization until further consideration.

NOTE: Due to some current dependency issues on my end that are still in the process of being resolved, these changes are only tested off of an older version of staging.  It is thus requested that further testing across multiple visualizable files be done by other users.

### DIFF
--- a/utils/pvacapi/controllers/database.py
+++ b/utils/pvacapi/controllers/database.py
@@ -151,62 +151,82 @@ def create_table(parentID, fileID, data, tablekey, db):
         rowid SERIAL PRIMARY KEY NOT NULL,\
         %s\
     )" % (tablekey, tablecolumns)
-    try:
-        with db.xact():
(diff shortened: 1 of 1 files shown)


susannasiebert:
I think it's ok to remove protein_position from being visualizable permanently. I can't imagine that it would ever be useful to anyone.

atwollam:
Looks like the tests have failed, so I'll have to wait until my dependency issues are resolved before I can test against this newer version of staging.

PR #423, opened 2019-07-09
Add manufacturability metrics to the all_epitopes file

### LINKED ISSUE #419: Add manufacturability metrics to the all_epitopes.tsv file
Doing so will make sure the metrics are reported in all of the downstream files.

susannasiebert:
Closes #419


susannasiebert:
Yeah, I was briefly considering changing the input file to be an intermediate file, but IDK, this seemed more straightforward.

PR #436, opened 2019-07-19
Switch to previous DB-file-read method if postgres user not superuser

atwollam:
Right now when reading in data from files to visualize, the postgres user must be a superuser.  This change allows for visualization when this is not the case, by switching to the previously used method for populating the database.

The second commit addresses the fact that the 'chromosome' column of files can occasionally be a string, rather than an integer (i.e. 'chr16' instead of 16).  This change temporarily adjusts the column type accordingly.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -759,7 +759,8 @@ class APITests(unittest.TestCase):
             self.assertEqual({mapping[key] for key in outputrow}-testrow.keys(), {'rowid'})
             del outputrow['rowid']
             for key in outputrow:
-                self.assertEqual(outputrow[key], parsedata(testrow[mapping[key]]), "Mismatch: %s"%key)
+                testval = testrow[mapping[key]] if key == 'chromosome' else parsedata(testrow[mapping[key]])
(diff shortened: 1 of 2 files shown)


utils/pvacapi/controllers/database.py
```diff
@@ -211,22 +264,26 @@ def create_table(parentID, fileID, data, tablekey, db):
...
             else:
-                print("Postgres could not access file.  Check to make sure that both the "
-                    "file and your current postgres user has the appropriate permissions.")
-                raise
-        else:
-            retry_transaction = False
+                #attempt to resolve by copying file to /tmp/, changing its permissions, and accessing it there
+                if file_permissions:
+                    print("InsufficientPrivilegeError raised in accessing file.\nAttempting workaround...")
+                    retry_transaction = True
```
susannasiebert: If you put everything from line 212-268 in its own method (making `file_permissions` a parameter), you can simply call that subroutine here with file_permission set to `False`. This would remove the whole `while` loop and the `retry_transaction` flag.

PR #438, opened 2019-07-24
Add new command `pvacfuse generate_protein_fasta`

### LINKED ISSUE #416: Add `pvacfuse generate_protein_fasta` command
Right now this subcommand only works for pVACseq but we should add it to the pVACfuse command tree as well. This might need some further development work to make it compatible with pVACfuse.

susannasiebert:
Closes #416


tools/pvacfuse/generate_protein_fasta.py
```diff
@@ -0,0 +1,131 @@
...
+        "peptide_sequence_length", type=int,
+        help="Length of the peptide sequence to use when creating the FASTA.",
+    )
+    parser.add_argument(
+        "output_file",
+        help="The output fasta file."
+    )
+    parser.add_argument(
+        "--input-tsv",
+        help = "A pVACseq all_epitopes or filtered TSV file with epitopes to use for subsetting the input file to peptides of interest. Only the peptide sequences for the epitopes in the TSV will be used when creating the FASTA."
```
susannasiebert:
```suggestion
        help = "A pVACfuse all_epitopes or filtered TSV file with epitopes to use for subsetting the input file to peptides of interest. Only the peptide sequences for the epitopes in the TSV will be used when creating the FASTA."
```

PR #445, opened 2019-08-06
Fix bugs in top score filter logic

susannasiebert:
- Bug in the filter method for NA expression values
- Handling of the ensembl ID for pVACfuse where the transcript might be
a combination
- Sorting of pVACbind to also use the Mutation column
- Tests should use the filecmp.cmp method to ensure order of output
files are correct after sorting

### DIFF
--- a/lib/top_score_filter.py
+++ b/lib/top_score_filter.py
@@ -78,13 +78,14 @@ class TopScoreFilter:
                                     line_with_max_expression = line_with_transcript_expression
                             filtered_lines.append(line_with_max_expression)
                         else:
-                            line_with_lowest_transcript_no = lines[0]
+                            line_with_lowest_transcript_id = lines[0]
+                            lowest_transcript_id = re.compile('ENST(\d+)').match(line_with_lowest_transcript_id['Transcript']).group(1)
                             for line in lines:
                                 transcript_id = re.compile('ENST(\d+)').match(line['Transcript']).group(1)
-                                lowest_transcript_id = re.compile('ENST(\d+)').match(line_with_lowest_transcript_no['Transcript']).group(1)
                                 if transcript_id < lowest_transcript_id:
-                                    line_with_lowest_transcript_no = line
-                            filtered_lines.append(line_with_lowest_transcript_no)
+                                    lowest_transcript_id = transcript_id
+                                    line_with_lowest_transcript_id = line
+                            filtered_lines.append(line_with_lowest_transcript_id)
 
             if self.file_type != 'pVACbind':
                 sorted_rows = lib.sort.default_sort(filtered_lines, self.top_score_metric)


lib/sort.py
```diff
@@ -11,11 +11,13 @@ def default_sort(rows, top_score_metric):
 
 def pvacbind_sort(rows, top_score_metric):
     if top_score_metric == 'median':
-        sorted_rows = sorted(rows, key=lambda row: (float(row['Best Score'])))
+        sorted_rows = sorted(rows, key=lambda row: (row['Mutation']))
+        sorted_rows = sorted(sorted_rows, key=lambda row: (float(row['Best Score'])))
         sorted_rows = sorted(sorted_rows, key=lambda row: (int(row['Sub-peptide Position'])))
         sorted_rows = sorted(sorted_rows, key=lambda row: (float(row['Median Score'])))
```
tmooney: Would it make sense to combine these into a single sort? Something like:
```python
sorted_rows = sorted(rows, key=lambda row: (float(row['Median Score']), int(row['Sub-peptide Position']), float(row['Best Score']), row['Mutation']))
```

susannasiebert: sure thing. Just pushed a commit with that update.

lib/top_score_filter.py
```diff
@@ -79,7 +80,9 @@ def execute(self):
                         else:
                             line_with_lowest_transcript_no = lines[0]
                             for line in lines:
-                                if float(line['Transcript'].replace('ENST','')) < float(line_with_lowest_transcript_no['Transcript'].replace('ENST','')):
+                                transcript_id = re.compile('ENST(\d+)').match(line['Transcript']).group(1)
+                                lowest_transcript_id = re.compile('ENST(\d+)').match(line_with_lowest_transcript_no['Transcript']).group(1)
```
tmooney: I'd be tempted to save the current lowest value instead of redetermining it on every loop iteration.  

(Do we still need the `float` casts that were in the previous version?  I'm not sure what they were protecting against here.)

susannasiebert: I wasn't sure whether an string comparison would yield different results then numerical comparisons (e.g.`'119' < '10'`) but I just tried it in a python console and it's doing the right thing without the float cast. 
I also made the change you suggested to save the current lowest value.

lib/top_score_filter.py
```diff
@@ -69,19 +70,22 @@ def execute(self):
...
-                            line_with_lowest_transcript_no = lines[0]
+                            line_with_lowest_transcript_id = lines[0]
+                            lowest_transcript_id = re.compile('ENST(\d+)').match(line_with_lowest_transcript_id['Transcript']).group(1)
                             for line in lines:
-                                if float(line['Transcript'].replace('ENST','')) < float(line_with_lowest_transcript_no['Transcript'].replace('ENST','')):
-                                    line_with_lowest_transcript_no = line
-                            filtered_lines.append(line_with_lowest_transcript_no)
+                                transcript_id = re.compile('ENST(\d+)').match(line['Transcript']).group(1)
+                                if transcript_id < lowest_transcript_id:
+                                    lowest_transcript_id = re.compile('ENST(\d+)').match(line['Transcript']).group(1)
```
tmooney: Seems like this could be:
```suggestion
                                    lowest_transcript_id = transcript_id
```

susannasiebert: Doh, of course.

Issue #447, opened 2019-08-13
Support multi-sample VCFs in the the `pvacseq generate_protein_fasta` command


susannasiebert:
Currently the `pvacseq generate_protein_fasta` command expects a single-sample VCF. However, we should add support for multi-sample VCFs to add feature parity to the `pvacseq run` command.

PR #450, opened 2019-08-14
Blast peptides against the reference proteome as a measure of self-similarity

### LINKED ISSUE #70: Add step to calculate self-similarity of MT epitope
Original issue: https://github.com/griffithlab/pVAC-Seq/issues/222

Take the peptide sequence and do some sort of similarity search against the ref proteome (BLAST) to see if the peptide naturally exists in the body. If so, it wouldn't be recognized as "foreign" antigen.

1. Implementation: Blast against the reference proteome
2. Implementation: Blast against the personal proteome

susannasiebert:
This PR adds a proteome-similarity match to the pVACseq/fuse pipelines. To do this we take the epitope and extract a longer peptide sequence around the mutation position and blast this sequence against the reference proteome. If there is a reference match of at least 8 amino acids, we count this as a reference match. This is output as true/false in the `Reference Match` column of the output reports.

As a side effect this PR creates a peptide fasta sequence in the output directory. This PR also adds a `--species` parameter to the pipelines which, for the purpose of this feature, limits the taxon of the blast search but also is used to check the validity of the chosen input alleles. The `--species` parameter defaults to `human`. 

TODO:
- Actually run this step in the post-processor (probably need to mock this in the test for faster runtime)
- Make the match sequence length configurable
- Output a tsv with all the reference hits
- Include this step when running the the genereate_protein_fasta command
  - When running the generate_combined_fasta in lib/pipeline.py it currently creates the manufacturability metrics and will also run manufacturability metrics if generate_protein_fasta includes this step. To avoid this we need to make that part optional and not run it in generate_combined_fasta.

Closes #70 and #447


lib/calculate_reference_proteome_similarity.py
```diff
@@ -0,0 +1,95 @@
...
+                    peptide = self.extract_n_mer(records_dict[line['Index']], int(line['Sub-peptide Position']), int(line['Mutation Position']), len(mt_amino_acids))
+                result_handle = NCBIWWW.qblast("blastp", "refseq_protein", peptide, entrez_query="{} [Organism]".format(self.species_to_organism[self.species]))
+                reference_match = False
+                for blast_record in NCBIXML.parse(result_handle):
+                    if len(blast_record.alignments) > 0:
+                        for alignment in blast_record.alignments:
+                            for hsps in alignment.hsps:
+                                matches = re.split('\+| ', hsps.match)
+                                for match in matches:
+                                    if len(match) >= 8:
```
chrisamiller: Should this '8' be a parameter (or derived from some other input)?  Though we typically search for 8-11mers, I'm imagining the case where someone only searches for 11mers.  In that case, a length of 11 would be the smallest possible match.  The counterpoint would be that 8-mers are potentially presented by MHC and could theoretically cause problems, but maybe we have 8 as the default and let the user configure it if they want?  Or maybe someone wants to be super-stringent and include anything that matches even a 6-mer. Mostly thinking out loud here -  if it's complex to add, then probably not a huge issue at the moment, but if it's easy, passing it through might be useful?

susannasiebert: We can make it a parameter. It could either be a new parameter or we could use the lowest epitope length used in the pipeline.

lib/calculate_reference_proteome_similarity.py
```diff
@@ -0,0 +1,95 @@
...
+            records_dict = {x.id.replace('MT.', ''): str(x.seq) for x in filter(lambda x: x.id.startswith('MT.'), records)}
+        return records_dict
+
+    def extract_n_mer(self, full_peptide, subpeptide_position, mutation_position, mt_length):
+        flanking_sequence_length = 7
+        mt_start = (subpeptide_position-1) + (mutation_position-1)
+        start = mt_start - flanking_sequence_length
+        if start < 0:
+            start = 0
+        end = mt_start + mt_length + flanking_sequence_length
```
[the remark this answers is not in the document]
susannasiebert: I tested this and it works as intended.

susannasiebert:
There are a lot of merge requests on this branch so I made #512 which also cleans up the commit history. Closing in favor of #512

PR #453, opened 2019-08-20
Amend AGFusion exon parser to also accept comma-delimited files

### LINKED ISSUE #452: AGFusion exon files may be comma-delimited
Right now the parser expects these files to be tab-delimited. We need to add logic to accept either.

susannasiebert:
Closes #452

### DIFF
--- a/lib/input_file_converter.py
+++ b/lib/input_file_converter.py
@@ -475,29 +475,13 @@ class FusionInputConverter(InputFileConverter):
             sequence = full_sequence.replace('*', '')
             return (fusion_position, sequence)
 
-    def determine_exon_file_delimiter(self, file_name):
-        with open(file_name, 'r') as fh:
-            reader = csv.DictReader(fh, delimiter='\t')
-            for record in reader:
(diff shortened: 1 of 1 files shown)


susannasiebert:
@tmooney indeed it does. I pushed a commit to use it instead.

lib/input_file_converter.py
```diff
@@ -479,7 +479,9 @@ def parse_exon_file(self, input_file):
         exon_file = input_file.replace('_protein.fa', '.exons.txt')
         five_prime_positions, three_prime_positions = [], []
         with open(exon_file, 'r') as fh:
-            reader = csv.DictReader(fh, delimiter='\t')
+            dialect = csv.Sniffer().sniff(fh.read())
+            fh.seek(0)
+            reader = csv.DictReader(fh, delimiter=dialect.delimiter)
```
tmooney: Would it make sense to pass `dialect=dialect` here instead of pulling out only the delimiter?  Guess it depends how much we trust the sniffer :smile:

PR #456, opened 2019-09-09
Add some more thorough explanations from the paper to the docs

susannasiebert:


docs/pvacvector/run.rst
```diff
@@ -13,6 +13,17 @@ Usage
...
+as the lowest/best binding affinities across various epitope lengths against
+the multiple prediction algorithms we offer. We also recommend setting a higher
+binding threshold of 1000nM. By default, pVACvector runs on the median binding
+affinity with a 500nM binding threshold. Running on the lowest/best binding affinity
+ensures that if even just one of eight algorithms thinks a peptide is a good
+binder, the junction will be rejected. However, running pVACvector with more
+conservative parameters also increases the likelihood of pVACvector not being
+able to find a valid path. Therefore, running pVACvector in these multiple modes,
+and choosing the overall most conservative solution is recommended.
+
```
chrisamiller: Suggested edit: 
It may be necessary to explore the parameter space a bit when running pVACvector.  As binding predictions for some sites vary substantially across algorithms, the most conservative settings may result in no valid paths, often due to one "outlier" prediction.  Carefully choosing which predictors to run may help ameliorate this issue as well.

In general, setting a higher binding threshold (say, 1000nM) and using the median binding value will lead to greater possibility of a design, while more conservative settings of 500nM and lowest/best binding value will give more confidence that there are no junctional neoepitopes.  

Our current recommendation is to run pVACvector several different ways, and choose the path resulting from the most conservative set of parameters.

chrisamiller: This seems to be something ripe for automation in the future. (run pvacvector a few different ways, find most conservative solution).  Will need to be coupled with further performance improvements, that we've talked about in the past a little bit!

susannasiebert: Agreed.

PR #468, opened 2019-10-24
Fix bug in split_file function of net_chop.py and netmhc_stab.py

susannasiebert:
The bug would result in the second (and subsequent chunk) of the split
file having one entry removed, which would then also be missing from the
output file, effectively dropping epitope predictions from the filtered
files when there were more than 100 entries.

This commit also moved the method to lib/utils.py so that it isn't
duplicated.

### DIFF
--- a/lib/net_chop.py
+++ b/lib/net_chop.py
@@ -7,21 +7,10 @@ import re
 import os
 from time import sleep
 import collections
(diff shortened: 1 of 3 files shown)


chrisamiller:
+1

PR #471, opened 2019-11-12
Support frameshift variants that start at first position of transcript

susannasiebert:
Frameshift variants that impact the first amino acid of a transcript would throw an error on line 133 because `previous_result` was undefined.


chrisamiller:
+1  yay for fixing another edge case!

PR #492, opened 2020-02-11
Remove all pVACvector tmp files if -k flag is not set

### LINKED ISSUE #461: Fix pVACvector --keep-tmp-files
We the restructuring of the pVACvector output files, the `--keep-tmp-files` flag doesn't remove all of the intermediate files. We need to decide which files to keep no matter what and which ones can be deleted.

susannasiebert:
Closes #461

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ setup(
         'swagger-spec-validator==2.1.0',
         'jsonschema==2.6.0',
         'mock',
-        'vaxrank',
+        'vaxrank==1.0.2',
     ],
     package_data={
         'tools.pvacseq': pvacseq_data_files,


tools/pvacvector/run.py
```diff
@@ -454,8 +454,10 @@ def main(args_input=sys.argv[1:]):
         VectorVisualization(results_file, base_output_dir, args.spacers).draw()
 
     if not args.keep_tmp_files:
-        shutil.rmtree(os.path.join(base_output_dir, 'MHC_Class_I'), ignore_errors=True)
-        shutil.rmtree(os.path.join(base_output_dir, 'MHC_Class_II'), ignore_errors=True)
+        for subdirectory in range(tries+1):
```
tmooney: It looks like `tries` already gets incremented right before we exit the `while` loop (line 443). Is it necessary to do so again here?

PR #494, opened 2020-02-18
Add step to pVACvector to output the most likely nucleic acid sequence for the result peptide sequence

### LINKED ISSUE #294: Create helper tool to generate DNA sequences for the pvacvector peptide list
Currently pvactools outputs a list of peptides and their optimal order.  The trial workflow deals with creating the DNA construct as a separate exercise.  In some cases a DNA manufacturer may have their own process for determining the optimal DNA sequence for encoding a series of peptides.

Some groups may wish to take greater control over this aspect of the process.  Particularly for experimental (non-clinical) use of these DNA constructs.  

Maybe we could consider having an additional step that proposes the DNA sequence for the string of peptides that comes out of pvacvector?

susannasiebert:
This uses the [Emboss backtranseq](https://www.ebi.ac.uk/Tools/st/emboss_backtranseq/) tool which "reads a protein sequence and writes the nucleic acid sequence it is most likely to have come from."

It outputs a DNA fasta file next to the pVACvector peptide output fasta file.

Closes #294

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -109,6 +109,7 @@ setup(
         'jsonschema==2.6.0',
         'mock',
         'vaxrank==1.0.2',
+        'xmltramp2',
     ],
     package_data={
         'tools.pvacseq': pvacseq_data_files,


tools/pvacvector/run.py
```diff
@@ -369,6 +372,46 @@ def identify_problematic_peptides(Paths, seq_dict):
             problematic_start.add(node)
     return (problematic_start, problematic_end)
 
+def create_dna_backtranslation(results_file, dna_results_file):
+    record = SeqIO.read(results_file, 'fasta')
+    seq_num = record.id
+    peptide = str(record.seq)
+    params = {
+        u'email': 'susanna.kiwala@wustl.edu',
```
tmooney: Does it make sense for you to take personal responsibility for everyone using this tool?

susannasiebert: Ah no, I meant to test it without an email to see if it would work. If email is indeed required we should make a new one for this use case.

susannasiebert:
@chrisamiller @malachig @obigriffith: @tmooney pointed out [here](https://github.com/griffithlab/pVACtools/pull/494#discussion_r380987982) that I was using my personal wustl address to make the API requests. I confirmed that _a_ email address is required so if we want to make this tool we would need to create a generic email address. We could reuse help@pvactools.org but I'm not sure under which circumstances they would use this email. We don't want to run the risk of getting spammed by status emails from them although I haven't received an emails for the jobs that I have run so they might only use it to track users.

chrisamiller:
hrmm... The email address thing is thorny.  We don't want to be responsible for any missuse (or just stupid use) by other groups.

Perhaps this is enough to tip the scales in favor of forking/recreating the code we talked about yesterday and do the reverse translation internally, rather than hitting an API?

susannasiebert:
That's my thought as well. Do we just want to go with the first codons from the tool you posted yesterday (https://github.com/Eslam-Samir-Ragab/RTCU/blob/master/additional/original_codon_usage.txt)?

susannasiebert:
Closing in favor of #495

PR #496, opened 2020-02-20
Add shorthands to run all available algorithms

### LINKED ISSUE #486: add "all" shorthand to run all available algorithms in pvacseq
Since this is a very common use case, I think it'd be handy.  I think the upsides (easier use) outweigh the potential downsides (discourages people from thinking carefully about their algorithm choices)

susannasiebert:
This PR adds three new shorthand options for prediction algorithms when running the individual tools. `all` will run all prediction algorithms,`all_class_i` will run all MHC Class I algorithms, and `all_class_ii` will run all MHC Class II algorithms.

This PR also moves some shared functionality to a `run_utils` module. This inadvertently revealed that the combined report bug fixed in #358 was never applied to pVACbind and pVACfuse. This PR fixes that bug for those two tools and updates the test data accordingly.

Closes #486

### DIFF
--- a/lib/run_utils.py
+++ b/lib/run_utils.py
@@ -4,25 +4,24 @@ import csv
 from lib.prediction_class import *
 
 def split_algorithms(prediction_algorithms):
-    class_i_prediction_algorithms = []
-    class_ii_prediction_algorithms = []
     if 'all' in prediction_algorithms:
-        class_i_prediction_algorithms = MHCI.prediction_methods()
-        class_ii_prediction_algorithms = MHCII.prediction_methods()
-    elif 'all_class_i' in prediction_algorithms or 'all_class_ii' in prediction_algorithms:
-        if 'all_class_i' in prediction_algorithms:
-            class_i_prediction_algorithms = MHCI.prediction_methods()
-        if 'all_class_ii' in prediction_algorithms:
-            class_ii_prediction_algorithms = MHCII.prediction_methods()
-    else:
-        for prediction_algorithm in sorted(prediction_algorithms):
-            prediction_class = globals()[prediction_algorithm]
-            prediction_class_object = prediction_class()
-            if isinstance(prediction_class_object, MHCI):
-                class_i_prediction_algorithms.append(prediction_algorithm)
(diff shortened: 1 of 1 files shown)


lib/run_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    class_ii_prediction_algorithms = []
+    if 'all' in prediction_algorithms:
+        class_i_prediction_algorithms = MHCI.prediction_methods()
+        class_ii_prediction_algorithms = MHCII.prediction_methods()
+    elif 'all_class_i' in prediction_algorithms or 'all_class_ii' in prediction_algorithms:
+        if 'all_class_i' in prediction_algorithms:
+            class_i_prediction_algorithms = MHCI.prediction_methods()
+        if 'all_class_ii' in prediction_algorithms:
+            class_ii_prediction_algorithms = MHCII.prediction_methods()
+    else:
```
tmooney: What if I want to specify `all_class_i` plus a single Class II algorithm (or vice versa)? Should that be supported?

lib/prediction_class.py
```diff
@@ -132,6 +132,12 @@ def prediction_classes(cls):
     def prediction_methods(cls):
         return sorted([prediction_class.__name__ for prediction_class in cls.prediction_classes()])
 
+    @classmethod
+    def prediction_methods_with_all(cls):
+        methods = cls.prediction_methods()
+        methods.extend(['all', 'all_mhc_i', 'all_mhc_ii'])
```
tmooney: In run_utils.py, it's looking for `all_class_i` and `all_class_ii`.  Do these get converted from `all_mhc_i` and `all_mhc_ii` somehow?

chrisamiller: my vote is for "all_class_i" over "all_mhc_i", fwiw

lib/run_utils.py
chrisamiller: maybe an easy way would be to create an array of prediction methods, expand the "all" options out to all their components, and then unique the array?  That then handles the case where you specify all and then list some others explicitly

PR #512, opened 2020-03-13
Blast peptides against the reference proteome as a measure of self-similarity (rebased)

### LINKED ISSUE #70: Add step to calculate self-similarity of MT epitope

susannasiebert:
This is a cleaned up version of #450 with out merge conflicts. It supersedes #450.

Closes #70 

TODO (this will be a separate issue): Calculate metrics when running `generate_protein_fasta` commands.

### DIFF
--- a/lib/run_argument_parser.py
+++ b/lib/run_argument_parser.py
@@ -43,12 +43,6 @@ class RunArgumentParser(metaclass=ABCMeta):
                  + "Required for Class I prediction algorithms.",
         )
         parser.add_argument(
-            "--species",
-            help="The species of the test subject(s). Will be used to check allele validity and for reference proteome search.",
-            default="human",
-            choices=['human', 'mouse', 'chimpanzee', 'macaque', 'cow', 'horse', 'pig'],
-        )
-        parser.add_argument(
             "--iedb-install-directory",
             help="Directory that contains the local installation of IEDB MHC I and/or MHC II."
         )
--- a/lib/run_utils.py
+++ b/lib/run_utils.py
@@ -23,20 +23,27 @@ def split_algorithms(prediction_algorithms):
             class_ii_prediction_algorithms.add(prediction_algorithm)
     return (sorted(list(class_i_prediction_algorithms)), sorted(list(class_ii_prediction_algorithms)))
 
-def split_alleles(alleles, species):
+def split_alleles(alleles):
     class_i_alleles = []
     class_ii_alleles = []
+    species = None
     for allele in sorted(set(alleles)):
         valid = 0
-        if allele in MHCI.all_valid_allele_names() and species == PredictionClass.species_for_allele(allele):
+        if allele in MHCI.all_valid_allele_names():
             class_i_alleles.append(allele)
             valid = 1
-        if allele in MHCII.all_valid_allele_names() and species == PredictionClass.species_for_allele(allele):
+        if allele in MHCII.all_valid_allele_names():
             class_ii_alleles.append(allele)
             valid = 1
         if not valid:
             print("Allele %s not valid. Skipping." % allele)
-    return (class_i_alleles, class_ii_alleles)
+        else:
+            allele_species = PredictionClass.species_for_allele(allele)
+            if species is None:
+                species = allele_species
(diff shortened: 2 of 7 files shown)


lib/calculate_reference_proteome_similarity.py
```diff
@@ -0,0 +1,144 @@
...
+        if peptide_sequence_length%2 == 0:
+            flanking_sequence_length = int(peptide_sequence_length/2)
+        else:
+            flanking_sequence_length = int((peptide_sequence_length-1)/2)
+        start = subpeptide_position - 1 - flanking_sequence_length
+        if start < 0:
+            start = 0
+        #This catches cases where the start position would cause too many leading wildtype amino acids, which would result
+        #in false-positive reference matches
+        diff_position = [i for i in range(len(wt_peptide)) if wt_peptide[i] != full_peptide[i]][0]
```
chrisamiller: as a non-pythonista, this syntax is a little confusing. I _think_ it's returning a list of the positions where the WT and MT peptides differ, right?

lib/prediction_class.py
```diff
@@ -370,6 +387,13 @@ class MHCII(PredictionClass, metaclass=ABCMeta):
     def needs_epitope_length(self):
         return False
 
+    @classmethod
+    def species_for_allele(self, allele):
+        if allele.startswith('H2'):
+            return 'mouse'
+        else:
+            return 'human'
+
```
chrisamiller: Is the difference between this and the above that no Class II algorithms support anything but human or mouse?  Would it be better to make a comprehensive list of class I and II prefixes in a dictionary (based on the valid_alleles list) and then we'd only have one function to update as new tools come online?

lib/run_argument_parser.py
```diff
@@ -42,6 +42,12 @@ def __init__(self, tool_name, input_file_help):
                  + "Typical epitope lengths vary between 8-11. "
                  + "Required for Class I prediction algorithms.",
         )
+        parser.add_argument(
+            "--species",
+            help="The species of the test subject(s). Will be used to check allele validity and for reference proteome search.",
+            default="human",
+            choices=['human', 'mouse', 'chimpanzee', 'macaque', 'cow', 'horse', 'pig'],
+        )
```
chrisamiller: Are the alleles themselves insufficient to infer the species?

lib/calculate_reference_proteome_similarity.py
susannasiebert: `i for i in range(len(wt_peptide)) if wt_peptide[i] != full_peptide[i]]` returns a list of positions that are different, and the subsequent `[0]` grabs the first entry, i.e. the first position that is different.

lib/run_argument_parser.py
susannasiebert: This is to prevent a user from using a combination of alleles from different species. .

lib/prediction_class.py
susannasiebert: For class I mouse alleles start with `H-2` and for class II they start with `H2`. We can still combined them, I think.

lib/run_argument_parser.py
chrisamiller: I guess what I'm asking is, does it make sense to simplify the inputs? Is it possible to just require the alleles, infer the species (and we can still check to make sure there is not more than one species represented and throw an error).  Seems like that would be a usability win - one less place for the user to screw up an input - but would it be a big hurdle in the code?

lib/prediction_class.py
susannasiebert: I just pushed a commit to consolidate these two methods.

lib/run_argument_parser.py
susannasiebert: I pushed a commit to infer the species from the alleles provided and error out if there is a mismatch.

PR #520, opened 2020-03-23
Allow users to specify a separate list of class II epitope lengths

### LINKED ISSUE #235: Better support for peptide lengths for class II prediction
Currently class II prediction appears limited to peptides of length 15 only.  This does not reflect the biology of MHC class II binding or even the capabilities of the class II predictors.  It appears that IEDB simply made the choice to only support length of 15?

https://help.iedb.org/hc/en-us/articles/114094152411-Why-does-the-MHC-class-II-binding-prediction-operate-only-on-15-mers-

We should confirm that the IEDB interface really does still have this limitation.  

If it does, and given that is has already been two years since they stated it would be fixed, we may want to consider a work around?

susannasiebert:
Closes #235 

Reviewers, please take special note of the new arguments in `lib/run_argument_parser.py` (https://github.com/griffithlab/pVACtools/compare/staging...epitope_length2?expand=1#diff-a06dc88b532df643a77a13749169da4aL38). Let me know if the chosen argument names work.


lib/prediction_class.py
```diff
@@ -396,7 +395,7 @@ def valid_allele_names(self):
         return self.valid_allele_names_for_class('class_ii')
 
     def valid_lengths_for_allele(self, allele):
-        return [15]
+        return [11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29,30]
```
tmooney: Could possibly use `range` here if listing numbers gets unwieldy.

lib/prediction_class.py
```diff
@@ -432,7 +431,7 @@ def valid_allele_names(self):
 
     def iedb_executable_params(self, iedb_executable_path, method, allele, input_file, epitope_length):
         allele = allele.replace('-DPB', '/DPB').replace('-DQB', '/DQB')
-        return "{} {} {} {}".format(iedb_executable_path, method, allele, input_file)
+        return "{} {} {} {} {}".format(iedb_executable_path, method, allele, input_file, epitope_length)
```
tmooney: Over on the class I side, this is:
https://github.com/griffithlab/pVACtools/blob/9867b5ee04c9aa853ea140485e14727ea84387f1/lib/prediction_class.py#L355-L356

Should this use the `str(epitope_length)`, too?  (Kind of surprising to me that the ordering is different, as well.)

susannasiebert: the `str` part is actually not needed (`format` will cast it for you). I believe this is a leftover from before `format` was a thing in python and I never removed it when I switched this to using `format`. 

The epitope_length addition on Class II is a new thing that was added to IEDB only recently so they appended that argument. It used to just default to 15. They probably didn't want to break backward compatibility. For Class II it also isn't required (it defaults to the old "standard" 15). Meanwhile it was always available for Class I so it got a more preferable position and there it is required.

tmooney: Thanks for the explanation!  I figured `str` couldn't be that important here or the tests would fail horribly 😄

chrisamiller:
+1

PR #524, opened 2020-03-24
Use the epitope length to determine the flanking sequence length during fasta generation

susannasiebert:


lib/pipeline.py
```diff
@@ -261,32 +260,37 @@ def generate_fasta(self, chunks):
...
-                'peptide_sequence_length'   : self.peptide_sequence_length,
                 'downstream_sequence_length': self.downstream_sequence_length,
                 'proximal_variants_file'    : self.proximal_variants_file,
             }
-            split_fasta_file_path = "%s_%s" % (self.split_fasta_basename(), fasta_chunk)
-            if os.path.exists(split_fasta_file_path):
-                status_message("Split FASTA file for Entries %s already exists. Skipping." % (fasta_chunk))
-                continue
             if self.input_file_type == 'pvacvector_input_fasta':
+                split_fasta_file_path = os.path.join(self.tmp_dir, "{}.fa.split_{}".format(self.sample_name, fasta_chunk))
```
tmooney: This version ended up copy+pasted a few times in this file.  If it was worth making a `split_fasta_basename()` to begin with, is it also worthwhile to have a version that works for this path (either via an option or a separate `pvacvector_split_fasta_basename()` or the like?

lib/proximal_variant.py
```diff
@@ -6,7 +6,8 @@
 import lib.utils
 
 class ProximalVariant:
-    def __init__(self, proximal_variants_vcf, pass_only):
+    #flanking_sequence_length is the number of bases (not amino acids!) to search on each side of a variant position
```
tmooney: If it's sometimes bases and sometimes amino acids that kind of makes me want it to have a different variable name for one of the cases!

lib/pipeline.py
susannasiebert: good idea. I made that change.

lib/proximal_variant.py
susannasiebert: I renamed this option to `flanking_bases` for clarity.

tools/pvacseq/generate_protein_fasta.py
```diff
@@ -64,11 +68,19 @@ def convert_vcf(input_vcf, temp_dir, sample_name):
     }
     if sample_name is not None:
         convert_params['sample_name'] = sample_name
+    if phased_proximal_variants_vcf is not None:
+        convert_params['proximal_variants_vcf'] = phased_proximal_variants_vcf
+        proximal_variants_tsv = os.path.join(temp_dir, 'proximal_variants.tsv')
+        convert_params['proximal_variants_tsv'] = proximal_variants_tsv
+        convert_params['flanking_bases'] = flanking_sequence_length * 4
```
riederd: convert_params['flanking_bases'] = flanking_sequence_length * 4
I do ask myself if this is correct? Shouldn't it be  "flanking_sequence_length * 3" ?
An AA is encoded by 3 bases, right?

susannasiebert: Yes, we are just giving ourselves a bit of buffer here.

riederd: OK.

PR #526, opened 2020-03-24
update citations, add link from docs back to source code, add link to…

malachig:
… dockerhub page


malachig:
Thanks @susannasiebert !

PR #532, opened 2020-03-27
Remove dependency on VaccinePeptideBase to allow for compatibility with all vaxrank versions

### LINKED ISSUE #493: Upgrade to latest vaxrank version
The current vaxrank version (1.2.0) is incompatible with our current usage in pVACtools:
```
  File "/home/travis/build/griffithlab/pVACtools/lib/calculate_manufacturability.py", line 8, in __new__
    return VaccinePeptideBase.__new__(
NameError: name 'VaccinePeptideBase' is not defined
```
I pinned vaxrank to the latest know working version for now but we should upgrade, if possible.

susannasiebert:
Closes #493

### DIFF
--- a/lib/calculate_manufacturability.py
+++ b/lib/calculate_manufacturability.py
@@ -1,20 +1,6 @@
 import csv
 from vaxrank.manufacturability import ManufacturabilityScores
-from vaxrank.vaccine_peptide import *
 from Bio import SeqIO
-
-class PvacpeptideVaccinePeptide(VaccinePeptide):
-    def __new__(cls, peptide):
-        return VaccinePeptideBase.__new__(
-            cls,
-            mutant_protein_fragment="",
-            mutant_epitope_predictions="",
-            wildtype_epitope_predictions="",
-            mutant_epitope_score="",
-            wildtype_epitope_score="",
-            num_mutant_epitopes_to_keep="",
-            manufacturability_scores=ManufacturabilityScores.from_amino_acids(peptide)
-        )
 
 class CalculateManufacturability:
     def __init__(self, input_file, output_file, file_type='pVACseq'):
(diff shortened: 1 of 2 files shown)


Stikus:
Hello, if I understand correctly - this merge and closing mentioned issue means that this [line](https://github.com/griffithlab/pVACtools/blob/master/setup.py#L111) `'vaxrank==1.0.2',` should lose version pinning?
For now, it is still causing problems with `isovar` requirements (fixed in latest releases).

susannasiebert:
This PR is for the upcoming version 2.0. It has not been incorporated in recent hotfix releases for version 1.5. What sort of dependency issues are you running into with isovar?

Stikus:
This is our error:
```
ERROR: isovar 0.9.0 has requirement pysam==0.9.0, but you'll have pysam 0.16.0.1 which is incompatible.
```
Reason is `vaxrank==1.0.2` [needs](https://github.com/openvax/vaxrank/blob/ba9d7196ea3733a59865b15bb5872297f55de383/setup.py#L67) `isovar>=0.8.5,<1.0.0` which needs `pysam==0.9.0` but we're installing `pysam`. Should we pin `pysam` version until 2.0 release?

susannasiebert:
Does this error prevent installation or does installation finish regardless? I ask because for me, I usually get this message but installation finishes successfully and I don't get any errors during runs. Regardless, I can pin pysam in the next hotfix release.

Stikus:
Everything working - we have this error for years :) If you think that pinning is good - go for it, but i'm not sure that `pysam==0.9.0` is ok for modern packages.

serge2016:
Maybe it is better to update `vaxrank` in hotfix releases for version 1.5?

susannasiebert:
I would need to apply this PR as a hotfix but that shouldn't be a problem.

serge2016:
Thank you, Susanna!

Issue #485, opened 2020-01-08
Create new vaccine selection report


chrisamiller:
Mockup and rough R code are attached.  Key features:
- "binning" of peptides into predefined categories/tiers
- revision that exposes more data in a more interpretable manner

As discussed, step 1 is to create a simple TSV like this as a report, for import into excel or something.

Step 2 will be to create an interactive HTML/JS page that adds the following features (in order of priority):
1) color coding of cell backgrounds
2) highlighting of mutant peptides in red
3) checkboxes that allow peptides to be selected, and an "export" button that spits out a tsv with only the selected peptides. 
4) "pop-up" boxes that allow a user to drill down into the data.
  - the first priority would allow for clicking on the number of peptides, and show all passing peptides for that mutation (along with transcript id, mutation position highlighted, and ic50 scores)

[example_ranking.xlsx](https://github.com/griffithlab/pVACtools/files/4036905/example_ranking.xlsx)

[ranking.R.gz](https://github.com/griffithlab/pVACtools/files/4036909/ranking.R.gz)
(wouldn't allow me to attach an R file, so I gzipped it)

malachig:
FYI, I am aware of at least two other ongoing efforts with partially overlapping goals.  Particularly for the visualization part.  Zach has been working on a Shiny app that does some of this for data from the HuMO project and Huiming has been working on a Shiny app that is meant to aid in selection of vaccine candidate in an immunotherapy tumor board setting.

susannasiebert:
I think we should have a followup meeting to discuss the future of pVACviz. One thing discussed yesterday was to replace the existing server-client model with a step in the pipeline that would produce html pages for the reports that could then be loaded in a browser. This would be easier maintainable and more robust than the existing framework.

malachig:
One thing for us to remember is that since this script goes back to the unfiltered all_epitopes file (which it must to get a sense of the collection of neoantigen candidates for each variant) ...  this means that it side steps a lot of what is happening in pVACseq run.  In a sense it deprecates much of what is happening in:

-  pvacseq binding_filter 
- pvacseq coverage_filter 
- pvacseq transcript_support_level_filter  
- pvacseq top_score_filter and 
- pvacseq generate_condensed_ranked_report

[...]

PR #540, opened 2020-04-06
Add step to create an aggregate report from the all_epitopes files

susannasiebert:
This step is added for pVACseq, pVACbind, and pVACfuse and is a python reimplementation of the R script in  #485 with some modification to make it work for pVACfuse and pVACbind.

It also adds this report generation as a standalone command to the tools' respective command line trees.

Areas for improvement that should be discussed:
- [ ] Use the --top-score-metric parameter to determine whether the median or lowest MT IC50 is used
- [ ] Add parameters to make some of the cutoffs/thresholds adjustable.


lib/aggregate_all_epitopes.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(self, input_file, output_file, file_type='pVACseq'):
+        self.input_file = input_file
+        self.output_file = output_file
+        self.file_type = file_type
+
+    #assign mutations to a "Classification" based on their favorability
+    def get_tier(self, mutation, vaf_clonal):
+        anchor_residue_pass = True
+        anchors = [1, 2, len(mutation["MT Epitope Seq"])-1, len(mutation["MT Epitope Seq"])]
+        if mutation["Mutation Position"] in anchors and mutation["Median WT Score"] < 1000:
```
chrisamiller, line 14: at some point, we're probably going to want to parameterize a lot of these (scores, expression levels, etc), but I think hardcoded is fine for this first version.

PR #542, opened 2020-04-10
Update main pVACtools workflow figure

### LINKED ISSUE #459: Update main workflow figure to include pVACbind

susannasiebert:
Closes #459


susannasiebert:
There aren't any current plans to expand pVACapi but we might want to add pVACbind there in the future as well. Not sure if that detail is important enough to update the figure for.

Josh has a versioning schema for any figures he makes. The rename reflects his version for the current figure.

PR #546, opened 2020-04-20
Set uniform permissions for all result files

### LINKED ISSUE #538: Consistent output file permissions
Occasionally, output files in the `combined/MHC_Class_I/MHC_Class_II` output directories have differing permissions, possibly due to some files being copied there from their initial locations. For consistency, pvactools could set these permissions. I think `644` or `444` are reasonable defaults.

susannasiebert:
Closes #538 

The owner needs execute permissions so that tests can run as expected (file compare on the output files throws a permissions error otherwise). Therefore, the permissions were bumped up to 744 instead of 644 (or 444) as suggested in #538.


tmooney:
Is the intent to restrict non-owners from viewing the results?  If not, directories will need the e`x`ecute bit for group and other as well. (i.e., `0o755`).

I bristle at setting the e`x`ecute bit on files that aren't meant to be executed--I'm not understanding why it needs to be set on TSVs?

PR #570, opened 2020-07-28
Remove conda dependency

susannasiebert:
IEDB 3.0 is python3 compatible so we will no longer need to use conda
in order to switch to a python2 env to run standalone IEDB.

### DIFF
--- a/tests/test_fasta_generator.py
+++ b/tests/test_fasta_generator.py
@@ -679,15 +679,14 @@ class FastaGeneratorTests(unittest.TestCase):
         self.assertEqual(os.path.getsize(generate_fasta_key_output_file.name), 0)
 
     def test_inframe_insertion_with_no_aa_change(self):
-        peptide_sequence_length        = '21'
         generate_fasta_input_file      = os.path.join(self.test_data_dir, 'input_no_aa_change.tsv')
         generate_fasta_output_file     = tempfile.NamedTemporaryFile()
         generate_fasta_key_output_file = tempfile.NamedTemporaryFile()
 
         generate_fasta_params = {
             'input_file'                : generate_fasta_input_file,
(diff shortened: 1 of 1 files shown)


lib/prediction_class.py
```diff
@@ -52,9 +52,9 @@ def check_length_valid_for_allele(self, length, allele):
 
     def predict(self, input_file, allele, epitope_length, iedb_executable_path, iedb_retries):
         if iedb_executable_path is not None:
-            arguments = self.iedb_executable_params(iedb_executable_path, self.iedb_prediction_method, allele, input_file, epitope_length)
+            arguments = ['python'].extend(self.iedb_executable_params(iedb_executable_path, self.iedb_prediction_method, allele, input_file, epitope_length))
```
tmooney, line 55: Would there be a benefit to using [`sys.executable`](https://docs.python.org/3/library/sys.html#sys.executable) here instead of trusting that `python` will be the correct interpreter?

lib/prediction_class.py
```diff
@@ -358,7 +358,7 @@ def check_length_valid_for_allele(self, length, allele):
             sys.exit("Length %s not valid for allele %s and method %s." % (length, allele, self.iedb_prediction_method))
 
     def iedb_executable_params(self, iedb_executable_path, method, allele, input_file, epitope_length):
-        return "{} {} {} {} {}".format(iedb_executable_path, method, allele, str(epitope_length), input_file)
+        return [iedb_executable_path, method, allele, str(epitope_length), input_file)]
```
tmooney, line 361:
```suggestion
        return [iedb_executable_path, method, allele, str(epitope_length), input_file]
```

lib/prediction_class.py
susannasiebert: That's a good idea I think.

PR #581, opened 2020-08-14
Fix bug in how we determine problematic peptides after running path-finding algorithm

susannasiebert:
Previously, we would use a rudimentary approach to determining problematic peptides after a failed path-finding attempt by simply calling `identify_problematic_peptides` again. This wouldn't actually return any results since this test first had to be passed to even make a path-finding attempt. As a result if a path-finding attempt was made but no path was found, then any subsequent trimmings of the peptides would be unsuccessful since no peptides were marked for trimming. Trimming would only happen if the pre-check before path finding failed.

This bugfix now tracks the "best" (but failing) path created by the annealing algorithm and returns all node-pairs of that path that don't have an edge (aka no good junction). It then returns those peptides that have a problematic end or start.

### DIFF
--- a/tools/pvacvector/run.py
+++ b/tools/pvacvector/run.py
@@ -296,10 +296,9 @@ def find_optimal_path(Paths, distance_matrix, seq_dict, seq_keys, base_output_di
     results_file = os.path.join(base_output_dir, args.sample_name + '_results.fa')
     with open(results_file, 'w') as f:
         name = list()
-        if Paths.has_edge(state[0], state[1]):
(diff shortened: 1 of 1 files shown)


susannasiebert:
Yeah, in the current implementation only the last spacer run matters. After the last spacer runs the problematic_start and problematic_end will be used to trim these problematic starts/ends from the peptide. The inner for loop over all the spacers then reruns with the trimmed peptides. In #580 this info will also be used to determine which junctions to add a spacer to (instead of adding spacers to all junctions).

PR #585, opened 2020-08-18
Handle prediction output with percentile None values

susannasiebert:


gschang:
+1, this is a runtime exception from external tools. This fixation will solve pvacvector job abortion with such exception.

PR #625, opened 2020-12-21
Set uniform permissions for all result files

susannasiebert:
This is a functional duplicate of https://github.com/griffithlab/pVACtools/pull/546 but with hotfix as the target branch for release in 1.5.13 because @guesu and @johnegarza were asking about it.


lib/run_utils.py
```diff
@@ -0,0 +1,11 @@
+import sys
+import os
+import csv
+from lib.prediction_class import *
```
tmooney, line 4: Could probably remove a few of these imports in this version.

tmooney (approved):
+1, looks like it'll get the job done!

gschang:
+1, Thank you so much for this hotfix, Susanna.

Issue #648, opened 2021-02-18
--netmhc-stab doesn't work


TinaGuo-hku:
Below is my commends to run a fasta file:
pvacbind run A0201.fasta Test HLA-A*02:01 MHCflurry MHCnuggetsI MHCnuggetsII NNalign NetMHC PickPocket SMM SMMPMBEC SMMalign A0201 -e1 9,10,11 --netmhc-stab

it works well on making binding predictions on allele, but when running NetMHCStabPan, there are some issues:

raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='www.cbs.dtu.dk', port=80): Max retries exceeded with url: /cgi-bin/webface2.fcgi (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fa6c4692390>: Failed to establish a new connection: [Errno 60] Operation timed out'))

I double checked my internet connect, it works well. 
Could you please help me out?

venkataddala:
I do have the same issue when i'm running with `--netmhc-stab --net-chop-method cterm --allele-specific-binding-thresholds --run-reference-proteome-similarity` and filtered.tsv file is blank only with headers

susannasiebert:
It looks like the NetMHCstabpan server was down last night but has since recovered. pVACtools uses that website to make stability predictions. Now that the site is back up, please try again and see if that resolves your issue.

venkataddala:
I ran again just now and NetMHCstab up and running on their site. I can see  all_epitopes.aggregated.tsv,  all_epitopes.tsv, but no sign of filtered.tsv.reference_matches file and blank filtered.tsv file

TinaGuo-hku:
Solved. Thank you.

TinaGuo-hku:
> I do have the same issue when i'm running with `--netmhc-stab --net-chop-method cterm --allele-specific-binding-thresholds --run-reference-proteome-similarity` and filtered.tsv file is blank only with headers

same thing here. All filtered.tsv files are blank with only headers.

susannasiebert:
[...]

PR #650, opened 2021-02-19
Update NetMHCstabpan and NetChop sserver addresses

### LINKED ISSUE #649: Migrate NetMHCstabpan and NetChop to use their new server
NetMHCstabpan is being migrated to https://services.healthtech.dtu.dk/service.php?NetMHCstabpan-1.0. NetChop is also moving to https://services.healthtech.dtu.dk/service.php?NetChop-3.1. I assume the old addresses will become unavailable eventually. We should get ahead of this and migrate our usage of the tools to their new addresses as well.

susannasiebert:
Closes #649 and #648

The old servers, even though they are still accessible, don't appear to currently return any results so this update is now a necessary bugfix.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -43,10 +43,10 @@ def make_response(data, files, path):
         reader.close()
         return response_obj
     else:
-        basefile = os.path.basename(data['configfile'])
+        configfile = data['configfile']
         reader = open(os.path.join(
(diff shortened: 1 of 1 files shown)


chrisamiller (approved):
+1 - seems like that new URL exists, at any rate!

Issue #660, opened 2021-04-16
Not enough values error in creating the aggregated epitopes report


nruthen:
* pvactools version: 2.0.0 (Using docker image griffithlab/pvactools:2.0.0)
* Python version: 3.7.0 (From python base image)
* Operating System: Debian GNU/Linux 10 (buster)

**Describe the bug**
I ran into an error similar to Issue #553 while running the pVACseq pipeline on VEP-annotated vcf files, however, this error occurred while the aggregated epitopes output was being generated.

**To Reproduce**
pVACtools command (via a bash script run within the docker container):

```
pvacseq run \
	-e1 8,9,10,11 \
	-e2 12,13,14,15,16,17,18 \
	--binding-threshold 1000 \
	--iedb-install-directory "/opt/iedb" \
	--normal-sample-name C57BL6_wes_1 \
	--normal-cov 1 \
	--normal-vaf 0.99 \
	--trna-cov 1 \
	--trna-vaf 0.01 \
	--tdna-cov 1 \
	--tdna-vaf 0.01 \
	--expn-val 1.0 \
	--maximum-transcript-support-level 5 \
	-t 16 \
	/cromwell_root/fc-ecef88af-5f5e-4d27-afe3-5c5f95ab8852/c4a4fbab-f2cd-4708-89a9-5721111fb2fd/mouse_variants2neoantigens/f49dcfd6-5251-4076-b443-449499242601/call-VEP/mutect_strelka_union_vep.vcf \
	GL261-8H3_wes_1 \
	H-2-Db,H-2-Kb \
	NetMHC NetMHCpan PickPocket \
	/cromwell_root
```
Input files: 
As the vcf and all_epitopes files are greater than 100 MB in size, perhaps we can share the files over Google Drive? 
I have attached the full log file to this report. 

**Log Output**
```
Combining Parsed Prediction Files
Completed
Creating aggregated report
Traceback (most recent call last):
  File "/usr/local/bin/pvacseq", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/tools/pvacseq/main.py", line 95, in main
    args[0].func.main(args[1])
  File "/usr/local/lib/python3.7/site-packages/tools/pvacseq/run.py", line 122, in main
    pipeline.execute()
  File "/usr/local/lib/python3.7/site-packages/lib/pipeline.py", line 504, in execute
    PostProcessor(**post_processing_params).execute()
[...]

PR #662, opened 2021-04-21
Update aggregate report creation to handle NAs correctly

susannasiebert:
Closes #https://github.com/griffithlab/pVACtools/issues/660

Previously, the way we were using pandas' read_csv function would result in a number of different string being interpreted as nulls. This includes `N/A` which is also a valid value for the `Mutation` column. This update will change the behavior to only interpret `NA` as nulls.


chrisamiller (approved):
+1

PR #667, opened 2021-04-29
Sleep 10 seconds after calling the BLAST API

susannasiebert:
Per [the BLAST API documentation](https://blast.ncbi.nlm.nih.gov/Blast.cgi?CMD=Web&PAGE_TYPE=BlastDocs&DOC_TYPE=DeveloperInfo) it is recommended to make only one call every 10 seconds.


chrisamiller (approved):
+1   

seems like local blast server support may need to be in the plan eventually here...

susannasiebert:
Oh definitely. I was planning on working on that next.

PR #668, opened 2021-05-04
Allow users to specifiy a blastp path for running reference proteome similarity

### LINKED ISSUE #516: Add option to run reference proteome similarity feature with standalone Blast
Right now we use bippython's blast wrapper for the reference proteome search. This is pretty slow. We should add the ability to use a standalone Blast installation. We would then also install Blast in our Docker container to use by default.

susannasiebert:
Closes #516


lib/calculate_reference_proteome_similarity.py
```diff
@@ -1,21 +1,30 @@
...
         self.input_fasta = input_fasta
         self.output_file = output_file
         self.metric_file = "{}.reference_matches".format(output_file)
         self.match_length = match_length
         self.species = species
         self.file_type = file_type
+        self.blastp_path = blastp_path
+        self.blastp_db = blastp_db
+        if self.blastp_db == 'refseq_select_prot' and self.species != 'human' and self.species != 'mouse':
+            raise Exception("refseq_select_prot blastp database is only compatible with human and mouse species.")
```
chrisamiller, line 27: Why is this restriction in place? Shouldn't the blastp algorithm be able to handle any species a user cares to build a database for?

susannasiebert: The `refseq_select_prot` database specifically is a limited set of reference proteins and only includes human and mouse. If someone wants to run other species they need to install the full reference proteome database (the `refseq_protein` option). I chose `refseq_select_prot` as the default since it will be faster than the full database for users running with the API.

susannasiebert: See https://ncbiinsights.ncbi.nlm.nih.gov/2020/10/16/blast-refseq-select/ for more info on the refseq select database.

PR #679, opened 2021-06-09
Issue669

nairod2000:
Pull request for issue 669


susannasiebert (approved):
This looks really good. I made one minor change suggestion but your solution works as well so feel free to ignore it if you like your implementation better. I'm still planning on doing some testing on my end before merging.

PR #711, opened 2021-09-28
Reporting el

nairod2000:
Added support for elution predictions with MHCflurry and NetMHCpan.

### DIFF
--- a/tests/test_data/pvacseq/tmpkgd5soto/MHC_Class_I/Test.tsv
+++ b/tests/test_data/pvacseq/tmpkgd5soto/MHC_Class_I/Test.tsv
@@ -1,1 +1,0 @@
-/Users/ssiebert/Documents/Work/pVACtools/tests/test_data/pvacseq/combine_and_condense/MHC_Class_I/Test.tsv
--- a/tests/test_data/pvacseq/tmpkgd5soto/MHC_Class_I/Test.tsv_1-24
+++ b/tests/test_data/pvacseq/tmpkgd5soto/MHC_Class_I/Test.tsv_1-24
@@ -1,1 +1,0 @@
-/Users/ssiebert/Documents/Work/pVACtools/tests/test_data/pvacseq/combine_and_condense/MHC_Class_I/Test.tsv_1-24
--- a/tests/test_data/pvacseq/tmpkgd5soto/MHC_Class_I/tmp
+++ b/tests/test_data/pvacseq/tmpkgd5soto/MHC_Class_I/tmp
@@ -1,1 +1,0 @@
-/Users/ssiebert/Documents/Work/pVACtools/tests/test_data/pvacseq/combine_and_condense/MHC_Class_I/tmp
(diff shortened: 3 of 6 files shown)


susannasiebert:
@nairod2000 I rebased this branch/PR on top of staging after merging in the PR that changed the full directory structure. You won't be able to just pull the PR since all of the commit hashed changed. The easiest would be to do a:
```
git fetch origin
git checkout reportingEL
git reset origin/reportingEL --hard
```
Make sure you don't have any uncommitted changes in your repo because the last command will delete those.

pvactools/lib/output_parser.py
```diff
@@ -433,6 +512,23 @@ def prediction_methods(self):
...
+                    row['MHCflurryEL Presentation %s' % suffix] = s
+                elif st == 'mhcflurry_processing_score':
+                    row['MHCflurryEL Processing %s' % suffix] = s
+                elif st == 'mhcflurry_presentation_percentile':
+                    row['MHCflurryEL Presentation %s' % suffix] = s
+                else:
+                    row['%s %s' % (pretty_method, suffix)] = s
+                
+        else:
+            # when would this ever get called?
```
susannasiebert, line 529: To answer this question: not all alleles/lengths are supported by all prediction algorithms. So you could have individual prediction algorithm score/percentile fields where no results are available for an output row for some of the prediction algorithms chosen by the user.

pvactools/lib/output_parser.py
```diff
@@ -433,6 +512,23 @@ def prediction_methods(self):
...
+            row['%s %s' % (prediction_method, suffix)] = 'NA'
```
susannasiebert, line 530:
```suggestion
            row['%s %s' % (pretty_method, suffix)] = 'NA'
```

pvactools/lib/output_parser.py
susannasiebert, line 529:
```suggestion
            # This covers cases where the allele/length of the result row is not supported by the prediction algorithm
```

nairod2000:
NetMHCpanEL and MHCflurry's EL based predictors were added in this pr. MHCflurry's EL based predictions include a presentation score, a processing score, and a presentation percentile. The processing score is a score that is computed from a model trained on an eluted ligand dataset (which is similar to netMHCpanEL). The presentation score and percentile are based off of an integrated model that is basing its prediction off of both processing and binding affinity. NetMHCpanEL's score, the processing score and the presentation score are all a value between 1 and 0, with higher scores indicating more favorable predictions, and thus cannot be included in the calculation of the summary metrics for scores (which are based on ic50 values). The percentiles for all methods are similarly calculated and thus the new percentiles are included in the summary metrics for percentiles.

New fields for the MHCflurryEL option are:
- "MHCflurryEL Presentation WT Score"
- "MHCflurryEL Presentation MT Score"
- "MHCflurryEL Processing WT Score"
- "MHCflurryEL Processing MT Score"
- "MHCflurryEL Presentation WT Percentile"
- "MHCflurryEL Presentation MT Percentile"

New fields for the NetMHCpanEL option are:
- "NetMHCpanEL WT Score"
- "NetMHCpanEL MT Score"
- "NetMHCpanEL WT Percentile"
- "NetMHCpanEL MT Percentile"

With support for arbitrary prediction methods, it might be good rename the
summary metric fields for "Best MT Score Method" and "Best MT Score" to
explicitly state that they are for ic50 predictions.

PR #734, opened 2021-11-04
Fix error with PASS filters not getting checked correctly after VCF parser module change

susannasiebert:
Gue Su noticed a problem with his recent trial runs of the 3.0.0rc3 docker image where the TSV files would be empty (https://github.com/griffithlab/immuno-planning/issues/3). This was caused by the switch the the VCFpy parser module which parses the FILTER column differently than PyVCF. PyVCF would parse a `PASS` filter into an empty array while VCFpy parses it into `['PASS']`. This update now checks for this value as well using the `--pass-only` argument.

### DIFF
--- a/pvactools/lib/input_file_converter.py
+++ b/pvactools/lib/input_file_converter.py
@@ -283,7 +283,7 @@ class VcfConverter(InputFileConverter):
                 continue
 
             filt = entry.FILTER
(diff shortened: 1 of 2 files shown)


gschang:
Thank you so much for this immediate attention. I will use the new version with clinical cases and update you this progress.

PR #745, opened 2021-12-03
pVACview bug fix

huimingx:
Fixed bug that was causing mismatch between export page and main table whenever regenerate table button was used. 
Removed load default button for non-demo version of the ui.R file for the shiny app.


susannasiebert (approved):
+1

PR #752, opened 2022-01-12
Fix sorting logic when lowest top score metric is chosen while running NetMHCstabpan

susannasiebert:
This fixes a syntax error introduced by #715 that would only be encountered when running NetMHCstabpan with the `lowest` top score metric.

This closes the error reported in https://github.com/griffithlab/immuno-planning/issues/6#issuecomment-1011440784


gschang:
Thank you so much for this work and help, Susanna.

PR #763, opened 2022-02-01
Enable filtering out non-pass variants when generating protein fasta standalone and for running NetChop

susannasiebert:
This will resolve an error @gschang encountered where NetChop would fail because the standalone fasta generation command, used to create the NetChop-specific FASTA file, would not respect the `--pass-only` flag used by the main pvacseq run command. Thus, extra entries were being created in the NetChop fasta file, which would result in the NetChop FASTA headers not matching up with the variant indexes of the all epitopes/filtered report files.

As a side effect, the standalone `pvacseq generate_protein_fasta` command now has a `--pass-only` option to skip variants with a FILTER.

### DIFF
--- a/pvactools/lib/net_chop.py
+++ b/pvactools/lib/net_chop.py
@@ -83,7 +83,7 @@ class NetChop:
                         index = line['Index']
                         epitope = line['MT Epitope Seq']
                     if index not in mt_records_dict:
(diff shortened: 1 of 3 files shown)


gschang:
Thanks you for this hotfix, Susanna. As updated, I tested the new `susannakiwala/pvactools:3.0.0rc8` release and confirmed that the bug is fixed now. All the pvacseq jobs with `3.0.0rc8` succeeded without this NetChop error.

Issue #606, opened 2020-11-04
Parameterize the aggregate report and add additional tiers


susannasiebert:
Currently the aggregate report uses a custom set of parameters/thresholds to sort epitope candidates into the different tiers. These should be parameterized, where applicable, and those parameters should be exposed to the user. In the pipeline, the parameters should then use the same thresholds as provided in the run command for the corresponding filters (e.g. binding threshold used for the binding filter could also be used for the Median MT Score threshold/tiering in the aggregate report). Additionally, some new tiers might be appropriate to correspond to the filters pVACseq applies, e.g. TSL.

Existing thresholds to parameterize:

Median WT Score (to determine anchor residues)
Tumor RNA VAF * Gene Expression (Pass)
Tumor RNA VAF * Gene Expression (Relaxed)
Median MT Score (Pass)
Median MT Score (Relaxed)
top score metric? (median vs lowest)
Current hard filters that aren’t used in aggregate tiering:

Fold Change
Allele specific MT Score filtering (if we use this the allele-specific value could be used for pass and double it could be used for relaxed)
Normal Depth
Normal VAF
Tumor RNA Depth
Tumor DNA Depth
Transcript Expression
Transcript support level
Notes:

~In the tiering Tumor DNA VAF is compared to an estimate of the clonal VAF instead of a specific numeric threshold~ This has been resolved with the new `--tumor-purity` parameter
In the tiering Tumor RNA VAF and Gene Expression are not compared individually to specific numeric thresholds; instead their product is compared
Should the relaxed MT Score threshold be a function of the pass MT Score threshold (e.g. double)?

susannasiebert:
@gschang also noted that currently the vaf clonal cutoff is hardcoded (i.e., it's calculated from the VAFs present in the sample). We will want to add a parameter so a custom, user-specified value is used as the cutoff instead of the calculated value.

susannasiebert:
#### Proposed parameterization
| Criteria | Tiering Hardcoded Cutoff | Run Default Cutoff | Proposed Parameterization |
|---|---|---|---|
| IC50 | Pass: 500<br />Relaxed: 1000 | --binding-threshold 500 | Pass: binding-threshold<br />Relaxed: binding-threshold x 2|
[...]

PR #811, opened 2022-06-07
Parameterize the binding threshold in the aggregate report

susannasiebert:
Part 1 of #606


susannasiebert:
I just pushed more commits to this PR to address the additional parameterization outlined in #606. So this closes #606

PR #822, opened 2022-07-06
Update mutation position column

### LINKED ISSUE #701: Update mutation position column or create new column with all mutated positions
Currently the mutation position column only shows where the mutation starts. In the case of frameshift or small indels, this does not show all the positions differing between the MT and WT peptide. 

A new approach would be to update the mutation position column or make a new column that lists or gives a range for the mutated aa in an epitope.

atwollam:
Resolves #701 and makes associated updates to incorporate the new format for the mutation position column.


susannasiebert (approved):
+1 I ran this change for HCC1395 and looked at it in pVACview and this all looks good to me.

PR #824, opened 2022-07-07
Only emit "Variant doesn't have protein position information" warning if the transcript consequence is supported in the first place

### LINKED ISSUE #823: Why so many no protein position variants
Thanks a lot for this great tool.

pvactools version: newest docker
Python version: 2.7
Operating System: ubuntu

I am working with VCF data called with GATK using hg19.
Now, when using VEP, I use the Grch37 reference

I run VEP with the following commands:

vep --input_file processed.vcf --format vcf --output_file vep.vcf --vcf --symbol --terms SO --tsl --hgvs --plugin Frameshift --plugin Wildtype --dir_cache /path  --offline

I get the output without errors.
Then when I run PVAC with:

pvacseq run SM.vcf SM HLA-A*02:01,HLA-B*35:01,HLA-C*07:02 NetMHCpan output/ -e1 9,10  

But all variants are report as "Variant doesn't have protein position information. Skipping.", but the variants including many missense mutation with hgvsp position, why pvactools can't identify?

thanks a lot.

susannasiebert:
This will cut down on noise since consequences like downstream_gene_variant and intron_variant (that can't be processed by pVACseq) don't have the protein position filled in and where clogging the stdout.

Closes #823

### DIFF
--- a/pvactools/lib/input_file_converter.py
+++ b/pvactools/lib/input_file_converter.py
@@ -305,6 +305,11 @@ class VcfConverter(InputFileConverter):
                     transcripts = self.csq_parser.parse_csq_entries_for_allele(entry.INFO['CSQ'], 'deletion')
 
                 for transcript in transcripts:
+                    transcript_name = transcript['Feature']
+                    consequence = self.resolve_consequence(transcript['Consequence'], reference, alt)
+                    if consequence is None:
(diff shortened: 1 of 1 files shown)


chrisamiller (approved):
+1

PR #847, opened 2022-10-05
Update image files to source from web

huimingx:
updated the source links for both the logos and anchor guide images to now source from github. I think this also means we can omit the step in pvacseq where it copies the images to a local copy?


huimingx:
Yes, I tried it after deleting the image files and it works the same I believe.

susannasiebert (approved):
+1

PR #852, opened 2022-10-13
Don't use pandas for filtering as it uses too much memory

### LINKED ISSUE #798: .all_epitopes.tsv result out, but process does not end for many hours
* pvactools version:docker
* Python version:3.7
* Operating System:centos

**Describe the bug**
the file *.all_epitopes.tsv result out in a early time, but process does not end for many hours, and the final result *.filtered.tsv does not come out

thanks a lot for your help

susannasiebert:
This is an alternative to #849 that abandons pandas for filtering altogether as it uses too much memory.

Closes #798


chrisamiller:
(unless I'm misunderstanding, I believe its an alternative to #849 , not 850)

chrisamiller (approved):
I didn't test it, but it looks correct, and is in some ways, easier to interpret than the pandas syntax, which is nice

susannasiebert:
> (unless I'm misunderstanding, I believe its an alternative to #849 , not 850)

you're correct. I fixed the original description. My apologies for the confusion

PR #879, opened 2022-11-25
to avoid calling read_input_file in the for loop in the execute method

yiolino:
This PR was created by modifying one part of aggregate_all_epitopes.py to make it faster.

When I run pVACbind, the issue was that if the input fasta file is large (tens of thousands of lines), the step to create the aggregated report is very time consuming.

When run with this implementation, a speedup of more than 10 times was achieved.

FYI, I am attaching the results of running pVACbind against a fasta file with 62700 lines.
Total: 14,500 s → 1,290 s (11.2 times faster）
aggregate_all_epitopes.py: 13,800 s → 510 s (27 times faster)

image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,13 +8,52 @@ on:
     branches: [ master, staging, hotfix]
 
 jobs:
-  tests:
-    name: Tests
+  test_python_3_6:
+    name: "Test Python 3.6"
+
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ['3.6']
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install system dependencies
+        run: |
+          sudo apt-get -qq update
+          sudo apt-get install -y ghostscript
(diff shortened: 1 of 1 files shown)


susannasiebert:
Thank you for submitting a pull request for this issue. We appreciate your contribution to our project. This is a great improvement. I originally made this change, because it was one of the suggested ways to reduce memory consumption as the aggregation step would routinely run out of memory for large dataset. Have you noticed any impact on memory usage with this change? This might be more of an issue for pVACseq than pVACbind since the pVACseq report has more columns and we also added additional memory saving measures as well. I would want to do some additional testing on my end to ensure no negative impact.

yiolino:
@susannasiebert
Thank you for your reply.

I understand that memory load was an issue, especially in the case of pVACseq.
When using pVACtools, memory usage was not monitored as we have never had a problem with memory. I will check my data as well to see the impact my changes are having.

yiolino:
@susannasiebert 
Although this is a result of pVACbind, I used memory_profiler to measure the change in memory load before and after modifying the code. There seems to be no significant change before and after the modification.
The attached figure shows the results of the measurement for the `execute` method of the `AggregateAllEpitopes` class in `aggregate_all_epitopes.py` (left: after code modification, right: original).
Next, I am planning to use pVACseq to see if there is any change.

image

susannasiebert (approved):
I tested this in pVACseq with one of our large samples as well and didn't run into any memory issues. +1

PR #911, opened 2023-03-06
Add new pVACsplice tool

mrichters:


pvactools/lib/load_ensembl_data.py
susannasiebert, line 1: Is this file being used in any module? Should it be deleted?

pvactools/lib/input_file_converter.py
```diff
@@ -455,6 +455,219 @@ def execute(self):
...
+                        'transcript_name'                : transcript_name,
+                        'transcript_support_level'       : tsl,
+                        'transcript_length'              : len(wildtype_amino_acid_sequence),
+                        'biotype'                        : biotype,
+                        'ensembl_gene_id'                : ensembl_gene_id,
+                        'hgvsc'                          : hgvsc,
+                        'hgvsp'                          : hgvsp,
+                        #'wildtype_amino_acid_sequence'   : wildtype_amino_acid_sequence,
+                        #'frameshift_amino_acid_sequence' : transcript['FrameshiftSequence'],
+                        #'fusion_amino_acid_sequence'     : '',
```
susannasiebert, line 635:
```suggestion
```

pvactools/lib/splice_pipeline.py
```diff
@@ -0,0 +1,193 @@
...
+
+    # creates transcripts.fa
+    def junction_to_fasta(self):
+        combined_df = self.combine_inputs()
+        if self.file_exists(self.create_file_path('fasta'), 'Junction fasta'):
+            pass
+        else:
+            print('Assembling tumor-specific splicing junctions')
+            for i in combined_df.index.to_list():
+                print(i)
```
susannasiebert, line 131:
```suggestion
```

pvactools/lib/junction_to_fasta.py
```diff
@@ -0,0 +1,220 @@
...
+        write_str = f'>WT.{self.fasta_index}\n{wt_seq}\n>ALT.{self.fasta_index}\n{alt_seq}\n'
+        if os.path.exists(self.output_file):
+            with open(self.output_file, 'r+') as f:
+                dup_content = re.search(write_str, f.read())
+                if dup_content == None:
+                    f.write(write_str)
+        else:
+            with open(self.output_file, 'w') as e:
+                e.write(write_str)
+        print("WT and ALT AA sequence written to file")
```
susannasiebert, line 200:
```suggestion
```

pvactools/lib/filter_regtools_results.py
```diff
@@ -0,0 +1,96 @@
...
+        # explode the transcript list and variant list
+        explode_junctions = filter_junctions.explode('transcripts', ignore_index=True).explode('variant_info', ignore_index=True).drop('index', axis=1) 
+        
+        explode_junctions = explode_junctions.rename(columns={'transcripts': 'transcript_id'})
+
+        return explode_junctions
+
+    def merge_and_write(self, pc_junctions, explode_junctions):
+        # merge dfs - explode_junctions from filtered file and pc_junctions from gtf file, all gtf info is based on tscript
+        # try:
```
susannasiebert, line 69:
```suggestion
```

pvactools/lib/filter_regtools_results.py
```diff
@@ -0,0 +1,96 @@
...
+        merged_df = explode_junctions.merge(pc_junctions, on='transcript_id').drop_duplicates()
+        # except KeyError:
+        #    sys.exit('The GTF dataframe is empty.')
```
susannasiebert, line 72:
```suggestion
```

pvactools/lib/filter_regtools_results.py
```diff
@@ -0,0 +1,96 @@
...
+        # drop repetitive or unneeded cols
+        merged_df = merged_df.drop(columns=['gene_names', 'gene_ids', 'variant_start', 'exon_number'])
+        # remove spaces from col names
+        # merged_df.columns = merged_df.columns.str.replace(r'\s+', '_', regex=True)
```
susannasiebert, line 76:
```suggestion
```

pvactools/lib/load_ensembl_data.py
mrichters: no, it can be deleted

susannasiebert (changes requested):

pvactools/lib/aggregate_all_epitopes.py
```diff
@@ -953,3 +951,75 @@ def get_tier(self, mutation, vaf_clonal):
...
+            'Num Passing Peptides': peptide_count,
+            'IC50 MT': best["{} IC50 Score".format(self.top_score_metric)],
+            '%ile MT': best["{} Percentile".format(self.top_score_metric)],
+            'RNA Expr': best["Gene Expression"],
+            'RNA VAF': best["Tumor RNA VAF"],
+            'RNA Depth': best["Tumor RNA Depth"],
+            'DNA VAF': best["Tumor DNA VAF"],
+            'Tier': tier,
+            'Evaluation': 'Pending',
+            'ID': key,
```
susannasiebert, line 1023: I think we can add the following entries here: `TSL`, `Allele Expr`, `Allele`, `Prob Pos`.
`Transcript` should be renamed to `Best Transcript`.

susannasiebert:
We need to make sure that the pVACsplice aggregated report works with the reference proteome feature. I don't think it will be a problem but it's something to double check. You can test this formally by either adding a new test to test_calculate_reference_proteome_similarity.py or by turning on this feature in a full pipeline test.

pvactools/lib/load_gtf_data.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def execute(self):
+
+        # make sure running pandas not polars
+        gtf_df_all = gtfparse.read_gtf(self.gtf_file, usecols=[
+            'feature', 'seqname', 'start', 'end', 'transcript_id',
+            'transcript_biotype', 'transcript_version', 'transcript_support_level', 'exon_number', 'gene_name',
+            'gene_id'], result_type='pandas')
+
+        gtf_df_all = gtf_df_all[~gtf_df_all['transcript_support_level'].isin(['', 'NA', 'Not Supported'])]
```
susannasiebert, line 22: I'm not sure that this is what we want. E.g., if someone runs GRCh37, there is no TSL data available and we would write 'Not Supported' to all entries. But then all transcripts would be included with this filtering step. I think 'Not Supported' is something we explicitly want to include.

pvactools/lib/load_gtf_data.py
```diff
@@ -0,0 +1,42 @@
...
+
+        gtf_df_all['transcript_support_level'] = gtf_df_all['transcript_support_level'].astype('int64')
+
+        # transcript and CDS - this is coding coordinates exon by exon (leaving out 5/3' UTRs in exon body)
+        gtf_df = gtf_df_all.loc[
+            (gtf_df_all['feature'].isin(['CDS', 'transcript'])) &
+            (gtf_df_all['transcript_support_level'] <= self.tsl) &
+            (gtf_df_all['transcript_biotype'] == 'protein_coding')
```
susannasiebert, line 30: Prefiltering on the `protein_coding` biotype is another thing we will want to discuss more. In pVACseq we only deprioritize entries that are from a not protein coding transcript but we don't downright exclude it.

pvactools/lib/load_gtf_data.py
susannasiebert: I went back and looked at some examples of Ensembl gtf files and had a couple additional thoughts:
- For GRCh37, the gtf file ([example](http://ftp.ensembl.org/pub/release-75/gtf/homo_sapiens/)) doesn't contain a "transcript_support_level" key. The `Not Supported` terminology is something we came up with for handling GRCh37 in pVACseq, not something straight from Ensembl. In pVACseq we check the build version of the VCF (see https://github.com/griffithlab/pVACtools/blob/staging/pvactools/lib/input_file_converter.py#L129) before parsing the TSL field. GTF files have a genome-version header that we could use for this purpose. If it's not GRCh38 we shouldn't do any of the TSL filtering.
- Looking at [GRCh38](http://ftp.ensembl.org/pub/release-106/gtf/homo_sapiens/), I see some transcript support level values like this: `"1 (assigned to previous version 3)"`. Does the gtfparse module already handle this or do we need to do some special parsing here? Or does the `astype` below end up converting this correctly?

pvactools/tools/pvacsplice/run.py
```diff
@@ -0,0 +1,239 @@
...
+    if args.fasta_size%2 != 0:
+        sys.exit("The fasta size needs to be an even number")
+    # gtf
+    if Path(args.gtf_file).suffix not in ['.gtf', '.tsv'] and not is_gz_file(args.gtf_file):
+        sys.exit('The gtf input path does not point to a gtf file.')
+    # vcf
+    if Path(args.annotated_vcf).suffix != '.vcf' and not is_gz_file(args.annotated_vcf):
+        sys.exit('The vcf input path does not point to a vcf file.')
+    # vcf gz.tbi index file
+    if not Path(f'{args.annotated_vcf}.tbi'):
```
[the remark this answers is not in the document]
mrichters: I changed this to `if Path(args.annotated_vcf).suffix != '.vcf' or not is_gz_file(args.annotated_vcf)`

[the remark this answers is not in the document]
mrichters: ok thanks I added the is_gz_file check to the .tbi line

pvactools/lib/load_gtf_data.py
mrichters: I do some special parsing to make sure there's no string in the final TSL col

mrichters: I'll double check about the Not Supported

mhoang22:
for the documentation for pvacsplice output file (output_file.rst) , in 'all_epitopes.tsv and filtered.tsv Report Columns' section,  I think we may want to change the column named 'Protein Positions' to 'Protein Position'.
I proposed this change because:
+ this column in the all_epitopes.tsv is 'Protein Position' (without an 's') -> we should delete the 's' to keep it consistent with the actual output
+ this column notes the position of the first amino acid of the selected epitopes. Not the protein positions that the mutant affects. -> we should delete the 's' to reflect that this position is for only 1 aa

susannasiebert (approved):
+1

PR #944, opened 2023-03-30
Various minor pVACview improvements

susannasiebert:
- Add a red border to Ref Match True cells
- Add some of the best peptide data to the Reference Matches detail tab so we don’t have to scroll up to the main table to cross reference the details like the amino acid change or gene
Screen Shot 2023-03-30 at 11 34 13 AM
- Add a Anchor Residue Fail column to the peptide detail table and add red border to peptides that fail this criteria
Screen Shot 2023-03-30 at 12 41 10 PM


susannasiebert:
Best Peptide and Query Peptide are now shown as graphs with color coding.

Screen Shot 2023-04-03 at 12 06 14 PM

PR #1020, opened 2023-09-03
Docker image to run pvacview on GCP

Layth17:


susannasiebert:
I was able to test this out and discovered a few minor issues with the update script, which I already talked to @Layth17 about. I'll note them here for posterity:
- If the docker build fails, the script will continue to run and look as if everything finished successfully. However, no new docker image is created and no new tagged cloud run is created
- If the update is run with an existing tag, the script finishes successfully but the existing tag won't be updated to point to the latest docker image.
Both of these situations should result in the script exiting with an appropriate error instead.

susannasiebert (approved):
+1

PR #1070, opened 2024-02-28
Remove BigMHC and DeepImmuno as direct dependencies

susannasiebert:
Unfortunately, PyPI does not allow you to upload a package with URL dependencies. All dependencies need to be other PyPI packages. As a workaround, this PR removes BigMHC and DeepImmuno as direct dependencies. Instead, users will need to manually install these two tools. This PR adds instruction to the docs on how to install these tools.


acoffman (approved):
+1

Issue #1088, opened 2024-04-04
Outputting accounted-for germline variants


evelyn-schmidt:
https://github.com/griffithlab/pVACtools/blob/7c4b383c554a62352c74a92884bc5214267810b1/pvactools/tools/pvacseq/generate_protein_fasta.py#L45C1-L48C6

The `generate_protein_fasta.py` script uses the `phased-proximal-variants-vcf` to create correct predicted sequences within the fasta file. It would be helpful if the list of variants that affect the final sequences were output to a file that could be easily viewed and eventually incorporated into the final neoantigen review files. 

As of now, during the manual review process of the neoantigen candidates, reviewers have to check to make sure that germline variants are correctly accounted for in the sequences (and a majority of the time they are). For clarity that these germline variants are being reviewed, a list of the accounted-for germline variants would be helpful.

susannasiebert:
This should be pretty straightforward. Such a file already gets created by the VCF conversion step so it just needs to be copied over from its temporary location.

PR #1091, opened 2024-04-09
Generate protein fasta prox vars

evelyn-schmidt:
Fix for issue [1088](https://github.com/griffithlab/pVACtools/issues/1088). Now the proximal_variants.tsv will be outputted along with the other files.

### DIFF
--- a/pvactools/tools/pvacseq/generate_protein_fasta.py
+++ b/pvactools/tools/pvacseq/generate_protein_fasta.py
@@ -194,7 +194,8 @@ def main(args_input = sys.argv[1:]):
     temp_dir = tempfile.mkdtemp()
     proximal_variants_tsv = convert_vcf(args.input_vcf, temp_dir, args.sample_name, args.phased_proximal_variants_vcf, args.flanking_sequence_length, args.pass_only)
     proximal_variants_file = "{}.proximal_variants.tsv".format(args.output_file)
-    shutil.copy(proximal_variants_tsv, proximal_variants_file)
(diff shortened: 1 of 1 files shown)


pvactools/tools/pvacseq/generate_protein_fasta.py
```diff
@@ -193,6 +193,12 @@ def main(args_input = sys.argv[1:]):
 
     temp_dir = tempfile.mkdtemp()
     proximal_variants_tsv = convert_vcf(args.input_vcf, temp_dir, args.sample_name, args.phased_proximal_variants_vcf, args.flanking_sequence_length, args.pass_only)
+    if proximal_variants_tsv is not None:
+        proximal_variants_file = "{}.proximal_variants.tsv".format(args.output_file)
+        with open(proximal_variants_tsv, 'r') as input_file:
+            with open(proximal_variants_file, 'w', newline='') as output_file:
+                output_file.write(input_file.read())
```
susannasiebert, line 200: While this works just fine, the more pytonic way to do this would be to use the shutil package. `shutil.copy(source_file_path, destination_file_path)` will copy a file from one location to another.

susannasiebert (changes requested):

pvactools/tools/pvacseq/generate_protein_fasta.py
evelyn-schmidt: I changed to this command and tested! Works great!

susannasiebert (approved):
+1

PR #1097, opened 2024-04-29
Update pvacseq_vignette.rst

chrisamiller:
pVACview doc edits

A few images and headers weren't appearing correctly, but I suspect that's an issue with the github preview, rather than something that's actually wrong.  May want to verify that, though!

### DIFF
--- a/docs/pvacview/pvacseq_module/pvacseq_vignette.rst
+++ b/docs/pvacview/pvacseq_module/pvacseq_vignette.rst
@@ -12,7 +12,7 @@
 
 Vignette
 ---------------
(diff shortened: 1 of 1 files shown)


mhoang22:
+ I have incorporated selected edits in the vignette rst in pvacview reviewer response branch.
+ I also fixed the headers names.
+ For figures, there are 3 places where the figures weren't displayed correctly in the preview (but displayed normal when building):
- 1 of which is the 'three-image-row'  
- 2 of which are 2 images named %ilePlot.png . I renamed the images to percentilePlot.png, that fixed the problem in the preview. 
Note: all of the edits mentioned are added to pvacview reviewer response branch, not master.

PR #1110, opened 2024-06-06
Added support for mouse specific allele anchor positions

ldhtnp:


pvactools/lib/top_score_filter.py
```diff
@@ -215,28 +228,14 @@ def find_best_line(self, lines):
...
-
     def is_anchor_residue_pass(self, line):
         if self.use_allele_specific_binding_thresholds:
             binding_threshold = self.allele_specific_binding_thresholds[line['HLA Allele']]
         else:
             binding_threshold = self.binding_threshold
 
         anchor_residue_pass = True
-        anchors = self.get_anchor_positions(line['HLA Allele'], len(line['MT Epitope Seq']))
+        anchors = get_anchor_positions(self, line['HLA Allele'], len(line['MT Epitope Seq']))
```
susannasiebert, line 238: I don't think we'll want to pass `self` here. In order to make this generic and future-proof, e.g. for other classes where the attributes might be named differently, we should pass along all attributes we use from `self` explicitly as their own parameters.

pvactools/lib/top_score_filter.py
```diff
@@ -137,6 +137,19 @@ def __init__(
...
+        mouse_anchor_positions = {}
+        for length in [8, 9, 10, 11]:
+            base_dir = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..'))
+            file_name = os.path.join(base_dir, 'tools', 'pvacview', 'data', "mouse_anchor_predictions_{}_mer.tsv".format(length))
+            values = {}
+            with open(file_name, 'r') as fh:
+                reader = csv.DictReader(fh, delimiter="\t")
+                for line in reader:
+                    allele = line.pop('Allele')
+                    values[allele] = line
```
susannasiebert, line 149: Another option here is to convert the strings into boolean values. You can do this via the `ast` library.
```
>>> import ast
>>> ast.literal_eval("False")
False
>>> type(ast.literal_eval("False"))
<class 'bool'>
```
```suggestion
                    values[allele] = [ast.literal_eval(v) for v in line]
```
However, this only works with "True"/"False", not "TRUE"/"FALSE" so you'd need to change the values in your TSV.

The advantage is that you can then change line 164 in `pvactools/lib/run_utils.py` to be `if val` instead of having to do a string comparison.

Nothing wrong with your current implementation; just wanted to mention an alternate approach.

susannasiebert (changes requested):

ldhtnp:
I am currently loading the mouse data from my github. This will need to be changed once the files are stored here

pvactools/tools/pvacview/anchor_and_helper_functions.R
```diff
@@ -75,10 +82,29 @@ peptide_coloring <- function(hla_allele, peptide_row) {
...
+    colors <- list()
+    for (i in 1:length(mouse_position_data)) {
+      if (mouse_position_data[i] == "True") {
+        colors <- append(colors, "blue")
+      } else {
+        colors <- append(colors, "lightblue")
+      }
+    }
+  } else {
+    if ((position == 1) || (position == 2) || (position == peptide_length-1) || (position == peptide_length)) {
```
susannasiebert, line 101: This could probably be simplified to something like:
```suggestion
    if (position %in% c(1, 2, peptide_length-1, peptide_length)) {
```

tests/test_data/run_utils/M_GC-OxParp_A-OxParp_A_FF_DNA.all_epitopes.tsv
susannasiebert, line 1: This file was probably committed accidentally.

susannasiebert: Disregard my comment. This looks correct.

susannasiebert (approved):
+1

PR #1116, opened 2024-06-20
fixing rounding issue in min-max slider

evelyn-schmidt:
@mhoang22 encountered a problem where the dynamic scatter doesn't show all candidates for the custom module. In her pvacsplice output, there were 5 candidate groups (corresponding to 5 genes associated with the neoantigens). She expected to see 5 dots corresponding to the 5 candidate groups. However, only saw 3 candidates. When examining the ‘Overview of Neoantigen Features’ board again, it looks like the 2 candidate groups with extreme VAF values (1st candidate: gene ATG12: tumor DNA VAF 0.432 , tumor RNA VAF 0.387; 5th candidate: gene TP53: tumor DNA VAF 0.852, tumor RNA VAF 0.998) are the missing candidates. If she natural-logged transformed both x and y axis, then she could see 5 candidates. She concluded that there was a reason that candidates with edge values were being excluded.

I found that these edge cases were being missed because of the rounding that happens to create the min-max slider. The range used for the min-max slider was originally rounded and now is just set as the dataset min/max. 

https://github.com/griffithlab/pVACtools/blob/efc101fa732262979f38edbadfb3c1060b6e6c2a/pvactools/tools/pvacview/server.R#L2120-L2121

So the change looks like this:
https://github.com/griffithlab/pVACtools/blob/ef27b63b9f209881e367e8de4d8d87b746620888/pvactools/tools/pvacview/server.R#L2117-L2118

### DIFF
--- a/pvactools/tools/pvacview/server.R
+++ b/pvactools/tools/pvacview/server.R
@@ -1607,8 +1607,8 @@ server <- shinyServer(function(input, output, session) {
 
           xvrbl_values <- df[[input$xvrbl]]
           range_values <- range(as.numeric(xvrbl_values), na.rm = TRUE)
-          min_value <- range_values[1]
-          max_value <- range_values[2]
+          min_value <- floor(range_values[1])
+          max_value <- ceiling(range_values[2])
 
 
           # Check if min_value and max_value are equal, set default values


susannasiebert (commented):
We could also consider using `floor` and `ceiling` in order to round down (for the min_value) and up (for the max_value), respectively. Not sure if, for display purposes, an int number is preferred, e.g. if the min_value and max_value are displayed on the chart axes.

evelyn-schmidt:
Yeah, @susannasiebert , I like the floor/ceiling suggestion because it looks a little cleaner on the slider. 

However, is the most clear option for users to set the max/min value to the actual min/max value in the dataset? So you can see the maximum and minimum values on the slider?

susannasiebert:
I personally don't think they need to match. They can always retrieve the actual min max value by hovering over the appropriate datapoint, right?

pvactools/tools/pvacview/server.R
```diff
@@ -1607,8 +1607,8 @@ server <- shinyServer(function(input, output, session) {
 
           xvrbl_values <- df[[input$xvrbl]]
           range_values <- range(as.numeric(xvrbl_values), na.rm = TRUE)
-          min_value <- as.numeric(format(round(range_values[1], 2), nsmall = 2))
-          max_value <- as.numeric(format(round(range_values[2], 2), nsmall = 2))
+          min_value <- range_values[1]
+          max_value <- range_values[2]
```
ldhtnp, line 1611:
```suggestion
          min_value <- range_values[1]
          max_value <- range_values[2]
```
```suggestion
          min_value <- floor(range_values[1])
          max_value <- ceiling(range_values[2])
```

pvactools/tools/pvacview/server.R
```diff
@@ -1655,9 +1655,9 @@ server <- shinyServer(function(input, output, session) {
 
         yvrbl_values <- df[[input$yvrbl]]
         range_values <- range(as.numeric(yvrbl_values), na.rm = TRUE)
-        min_value <- as.numeric(format(round(range_values[1], 2), nsmall = 2))
-        max_value <- as.numeric(format(round(range_values[2], 2), nsmall = 2))
-
+        
+        min_value <- range_values[1]
+        max_value <- range_values[2]
```
ldhtnp, line 1660:
```suggestion
        min_value <- range_values[1]
        max_value <- range_values[2]
```
```suggestion
        min_value <- floor(range_values[1])
        max_value <- ceiling(range_values[2])
```

pvactools/tools/pvacview/server.R
```diff
@@ -2042,9 +2038,8 @@ server <- shinyServer(function(input, output, session) {
           
           xvrbl_values <- df[[input$xvrbl_custom]]
           range_values <- range(as.numeric(xvrbl_values), na.rm = TRUE)
-          min_value <- as.numeric(format(round(range_values[1], 2), nsmall = 2))
-          max_value <- as.numeric(format(round(range_values[2], 2), nsmall = 2))
-          
+          min_value <- range_values[1]
+          max_value <- range_values[2]
```
ldhtnp, line 2042:
```suggestion
          min_value <- range_values[1]
          max_value <- range_values[2]
```
```suggestion
          min_value <- floor(range_values[1])
          max_value <- ceiling(range_values[2])
```

pvactools/tools/pvacview/server.R
```diff
@@ -2114,14 +2109,15 @@ server <- shinyServer(function(input, output, session) {
...
         
-        df <- df[is.finite(df[[input$yvrbl_custom]]),]
+        #df <- df[is.finite(df[[input$yvrbl_custom]]),]
         yvrbl_values <- df[[input$yvrbl_custom]]
         range_values <- range(as.numeric(yvrbl_values), na.rm = TRUE)
-        min_value <- as.numeric(format(round(range_values[1], 2), nsmall = 2))
-        max_value <- as.numeric(format(round(range_values[2], 2), nsmall = 2))
+        
+        min_value <- range_values[1]
+        max_value <- range_values[2]
```
ldhtnp, line 2117:
```suggestion
        min_value <- range_values[1]
        max_value <- range_values[2]
```
```suggestion
        min_value <- floor(range_values[1])
        max_value <- ceiling(range_values[2])
```

ldhtnp (changes requested):
Everything looks good to me. I agree with using floor and ceiling to have cleaner values displayed on the slider. I provided suggestions to add this

PR #1174, opened 2025-01-07
Prevent pVACvector from clipping Best Peptide

### LINKED ISSUE #1164: Prevent pVACvector from clipping epitopes
If a "core" epitope in a pVACvector input fasta sequence is located close to the start or end of the sequence, it could potentially get clipped. We should update pVACvector to prevent this. However, input fastas currently don't contain information about the location of the core epitope in the sequence. This update could be accomplish as follows:
- When creating a protein fasta using any of the `generate_protein_fasta` commands in combination with an input filtered.tsv of aggreated.tsv, mark the location of the epitopes from those tsvs in relation to the result protein sequence in the fasta headers.
- Use the information from the fasta headers to prevent pVACvector from clipping the core epitope.

susannasiebert:
When running the `pvacseq generate_protein_fasta` command in combination with an aggregated report as the `--input-tsv` the output fasta will now contain information about the Best Peptide for mutant fasta entries. This information is then used by pVACvector to prevent clipping into the Best Peptide.

Closes #1164

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
   run_tests:
     name: "Test pVACtools in various Python versions"
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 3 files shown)


chrisamiller (approved):
This looks great.  I briefly considered whether this best peptide check should be an option, instead of running all the time, but I'm struggling to come up with a scenario where we'd want to turn it off, so as is seems fine.  +1

Issue #457, opened 2019-09-16
Support multiple versions of prediction algorithms


susannasiebert:
IEDB has now added support for running multiple version of some prediction algorithms. We should leverage that functionality. I suggest that we extend the list of prediction algorithms to have on item per supported algorithm-version combination. This will break backwards compatibility since, for example, `NetMHC` would now be `NetMHC-4.0` and `NetMHC-3.4`. We could have `NetMHC` default to the latest but since that version might change in the future, this might not be as desirable as explicit versioning. We could also support both, I think.

malachig:
This seems important from a provenance and reproducibility perspective.  It would make it more transparent which algorithms are actually being used and would allow for more consistent runs across time for users that need locked down pipelines.  

Note that using the docker container release would keep things consistent but this change would still be an improvement in clarity.

susannasiebert:
Unfortunately, while the IEDB API supports multiple version of the prediction algorithms, the standalone package does not. I will remove this from the 2.0 milestone for now. Maybe we can revisit this in a future release.

susannasiebert:
At least for NetMHCIIpan, multiple versions are now supported in the latest standalone IEDB release (3.1.12) as well as the API. We should add support for specifying the NetMHCIIpan version.

PR #1181, opened 2025-01-22
Added argument for specifying NetMHCIIpan/NetMHCIIpanEL version

ldhtnp:
I added the argument `--netmhc_pan_version` to pvacseq, pvacbind, pvacfuse, pvacsplice, and pvacvector. The argument supports four versions: 4.0, 4.1, 4.2, and 4.3. I also added an argument that can be used to list the valid versions `valid_netmhcpan_versions`. Alongside the additions I had to make to get the arguments implemented, I had to modify the code used to remove the sample name and grab the method name in `output_parser.py` because the method name can now include a version number. Before this modification, the version number would be trimmed and dropped off of the method name resulting in issues downstream.

This is for issue [#457](https://github.com/griffithlab/pVACtools/issues/457#issuecomment-2391360664)


pvactools/lib/run_argument_parser.py
```diff
@@ -111,6 +111,12 @@ def __init__(self, tool_name, input_file_help):
             default=1,
             help="Number of threads to use for parallelizing peptide-MHC binding prediction calls.",
         )
+        parser.add_argument(
+            "--netmhc_pan_version",
+            choices=["4.3", "4.2", "4.1", "4.0"],
+            default=None,
```
susannasiebert, line 117: If you set the default here to "4.1" you can remove some of the logic that checks whether the netmhcpan version is set. 4.1 is the version that is already used.

pvactools/tools/pvacseq/run.py
```diff
@@ -74,6 +74,14 @@ def main(args_input = sys.argv[1:]):
     input_file_type = 'vcf'
     base_output_dir = os.path.abspath(args.output_dir)
 
+    if (args.netmhc_pan_version == '4.0' and args.iedb_install_directory is not None):
+        raise Exception("Standalone IEDB does not support version 4.0")
+
+    if (args.netmhc_pan_version and ("NetMHCIIpan" in args.prediction_algorithms or "NetMHCIIpanEL" in args.prediction_algorithms)):
+        NetMHCIIVersion.netmhcii_pan_version = args.netmhc_pan_version
+    elif (args.netmhc_pan_version and not ("NetMHCIIpan" in args.prediction_algorithms or "NetMHCIIpanEL" in args.prediction_algorithms or "all_class_ii" in args.prediction_algorithms)):
+        raise Exception("NetMHCIIpan and NetMHCIIpanEL version was specified but neither algorithm was selected")
```
susannasiebert, line 83: I think we can silently ignore this and don't need to error out. There is no adverse effect if it is set but not used. Additionally, this check would now always error when we add a default for the version, as suggested.

pvactools/lib/run_argument_parser.py
```diff
@@ -111,6 +111,12 @@ def __init__(self, tool_name, input_file_help):
             default=1,
             help="Number of threads to use for parallelizing peptide-MHC binding prediction calls.",
         )
+        parser.add_argument(
+            "--netmhc_pan_version",
```
susannasiebert, line 115: You'll want to name this argument `--netmhcii-pan-version` (or `--netmhc-ii-pan-version` or `--netmhciipan-version`; not sure which is better - might want to put it to a vote on Slack) to avoid confusion since there is a different class I algorithm NetMHCpan. Additionally, using dash instead of underscore is the standard in Python for variant names. The parser automatically changes the parsed variable to use underscores.

pvactools/lib/netmhc_pan_version.py
```diff
@@ -0,0 +1,28 @@
+import argparse
+
+
+class NetMHCPanVersion:
```
susannasiebert, line 4: You'll want to rename this class `NetMHCIIPanVersion` to avoid confusion with the `NetMHCpan` algorithm and rename classes/variables/arguments in other places as well to match.

susannasiebert (changes requested):
This looks pretty good overall. I made a few comments for suggested changes.

susannasiebert:
Also, you'll want to add tests for your new command in the `test_pvacseq|bind|vector|splice.py` files. Check out how the valid_alleles and valid_algorithms are tested in there and implement similar checks. That should probably get your coverage up as well and make the coveralls check pass.

pvactools/lib/output_parser.py
```diff
@@ -172,8 +172,11 @@ def get_scores(self, line, method):
             return {'score': float(line['BigMHC_IM'])}
         elif method.lower() == 'netmhcpan_el':
             return {'score': float(line['score'])}
-        elif method.lower() == 'netmhciipan_el':
-            return {'score': float(line['score'])}
+        elif 'netmhciipan_el' in method.lower():
+            try:
+                return {'score': float(line['score'])}
+            except:
+                return {'ic50': float(line['ic50'])}
```
susannasiebert, line 179: This block of code concerns me. NetMHCIIpan_el is elution algorithm and as such doesn't return an ic50. Is there a specific version combo that returns this header?

susannasiebert:
I think we are missing the standalone `valid_netmhciipan_versions` command from the pVACfuse tree. You will also want to add documentation for the new commands in each of the `docs/pvacseq|fuse|splice|bind|vector/additional_commands.rst` files.

susannasiebert:
The Features documentation pages for the various tools (e.g., https://pvactools.readthedocs.io/en/latest/pvacseq/features.html) has a table that list the version of each tool that is supported. We should add the additional NetMHCIIpan/NetMHCIIpanEL versions, note the default, as well as the one version that is IEDB API-only.

susannasiebert (approved):
+1

Issue #1176, opened 2025-01-07
Investigate mouse_anchor readTable warnings


malachig:
When pVACview is loaded in R the following warnings are presented

```
Warning in read.table(curl("https://raw.githubusercontent.com/griffithlab/pVACtools/2a44403d9844a27cfca05bb158b87b747aba3ae2/pvactools/tools/pvacview/data/mouse_anchor_predictions_8_mer.tsv"),  :
  incomplete final line found by readTableHeader on 'https://raw.githubusercontent.com/griffithlab/pVACtools/2a44403d9844a27cfca05bb158b87b747aba3ae2/pvactools/tools/pvacview/data/mouse_anchor_predictions_8_mer.tsv'
Warning in read.table(curl("https://raw.githubusercontent.com/griffithlab/pVACtools/2a44403d9844a27cfca05bb158b87b747aba3ae2/pvactools/tools/pvacview/data/mouse_anchor_predictions_11_mer.tsv"),  :
  incomplete final line found by readTableHeader on 'https://raw.githubusercontent.com/griffithlab/pVACtools/2a44403d9844a27cfca05bb158b87b747aba3ae2/pvactools/tools/pvacview/data/mouse_anchor_predictions_11_mer.tsv'
```

Perhaps an issue with the end of some of these files?

PR #1184, opened 2025-01-29
Fixed the mouse anchor data files, warnings no longer appear

ldhtnp:
Malachi noticed that two warnings would appear for the 8mer and 11mer mouse anchor prediction tsv files. I resolved the issue by remaking the two tsv files and adding a newline at the end of each. Because we are curling the data from github the warnings will still show till this fix is merged.

Resolves issue [#1176](https://github.com/griffithlab/pVACtools/issues/1176)


susannasiebert (changes requested):
Thank you for tackling this issue. 

Because pVACview references these files by commit hash you will also need to update the GitHub file paths in https://github.com/griffithlab/pVACtools/blob/master/pvactools/tools/pvacview/anchor_and_helper_functions.R to reference the new commit hash from this PR and then commit that change as well.

Also, I think this same issue also affects the 9mer and 10mer files so those will need fixing as well.

ldhtnp:
Similar to Malachi, I only got the warning with the 8mer and 11mer files. The 9mer and 10mer are fine on my end

susannasiebert:
I'm not sure why they wouldn't cause a warning but when I open the 9mer and 10mer files I don't see a return character on the last line. I think for consistency we should fix them as well.

susannasiebert (approved):
+1

Issue #1142, opened 2024-09-09
Allow passing candidates to be based on binding OR percentile cutoff.


malachig:
Based the current docs it sounds a candidate it filtered out unless it passes both of these cutoffs.  

```
  -b BINDING_THRESHOLD, --binding-threshold BINDING_THRESHOLD
                        Report only epitopes where the mutant allele has ic50
                        binding scores below this value. (default: 500)
  --percentile-threshold PERCENTILE_THRESHOLD
                        Report only epitopes where the mutant allele has a
                        percentile rank below this value. (default: None)
```

Is that a correct interpretation for what would happen if you set a value for both of these?

It would be nice if this could be interpreted as an either OR parameter.  So that for example, if a candidate is <500 nm OR <1% it would be considered a pass.

susannasiebert:
Yes, that is correct. Is that a general change to the behavior we want to make or something that should be configurable?

malachig:
It just keeps adding more complexity, but I guess it could be nice to have this configurable based on intent:

- One wishes to be conservative, so one requires BOTH criteria to be met to get a pass
- One wishes to get more candidates, so one required EITHER criteria to be met to get a pass

PR #1185, opened 2025-01-29
Behavior to allow different percentile threshold strategies for candidates

ldhtnp:
I added the parameter `--percentile-threshold-strategy` that currently does nothing. Further logic needs to be added to incorporate it downstream

Associated issue: [#1142](https://github.com/griffithlab/pVACtools/issues/1142)

### DIFF
--- a/docs/pvacbind/filter_commands.rst
+++ b/docs/pvacbind/filter_commands.rst
@@ -55,6 +55,12 @@ to ``lowest``, this threshold is applied to the ``Best Percentile`` column. When
 it is set to ``median``, the threshold is applied to the ``Median
 Percentile`` column.
 
+When the ``--percentile-threshold`` flag is set, the candidate inclusion strategy can be
+specified by using the ``--percentile-threshold-strategy`` parameter. The parameter has two
+options ``conservative`` (default) and ``exploratory``. The 'conservative' option requires a candidate 
+to pass BOTH the binding threshold and percentile threshold, while the 'exploratory' option requires
+a candidate to pass EITHER the binding threshold or percentile threshold.
+
 By default, entries with ``NA`` values will be included in the output. This
 behavior can be turned off by using the ``--exclude-NAs`` flag.
 
--- a/docs/pvacbind/output_files.rst
+++ b/docs/pvacbind/output_files.rst
@@ -234,6 +234,11 @@ provided to the pVACfuse run:
    * - ``--percentile-threshold``
      - When set, use this threshold to filter epitopes on the %ile MT score in addition to having to meet the binding threshold.
      - None
+   * - ``--percentile-threshold-strategy``
+     - Specify the candidate inclusion strategy. The ``conservative`` option requires a candidate to pass BOTH the binding threshold
+       and percentile threshold (if set). The ``exploratory`` option requires a candidate to pass EITHER the binding threshold or
+       the percentile threshold.
+     - conservative
 
 Tiers
 *****
@@ -259,7 +264,9 @@ Criteria Details
    * - Binding Criteria
      - Pass if Best Peptide is strong binder
      - ``IC50 MT < binding_threshold`` and ``%ile MT < percentile_threshold``
-       (if ``--percentile-threshold`` parameter is set)
+       (if ``--percentile-threshold`` parameter is set and 'conservative' ``--percentile-threshold-strategy`` is used) or
+       ``IC50 MT < binding_threshold`` or ``%ile MT < percentile_threshold``
+       (if 'exploratory' ``--percentile-threshold-strategy`` is used)
 
 
 .. _pvacbind_reference_matches:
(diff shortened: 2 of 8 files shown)


pvactools/lib/run_argument_parser.py
```diff
@@ -71,6 +71,13 @@ def __init__(self, tool_name, input_file_help):
             help="Report only epitopes where the mutant allele "
                  +"has a percentile rank below this value."
         )
+        parser.add_argument(
+            '--percentile-threshold-strategy',
+            choices=['conservative', 'exploratory'],
+            help="Specify the candidate inclusion strategy. Conservative requires a candidate to pass BOTH the binding threshold and percentile threshold."
+                 + " Exploratory requires a candidate to pass EITHER the binding threshold or the percentile threshold.",
```
susannasiebert, line 78:
```suggestion
            help="Specify the candidate inclusion strategy. The 'conservative' option requires a candidate to pass BOTH the binding threshold and percentile threshold (default)."
                 + " The 'exploratory' option requires a candidate to pass EITHER the binding threshold or the percentile threshold.",
```

pvactools/lib/aggregate_all_epitopes.py
```diff
@@ -452,6 +455,17 @@ def get_tier(self, mutation, vaf_clonal):
             if mutation["Transcript Support Level"] > self.maximum_transcript_support_level:
                 tsl_pass = False
 
+        ic50_pass = mutation["{} MT IC50 Score".format(self.mt_top_score_metric)] < binding_threshold
+        percentile_pass = (
+            self.percentile_threshold is not None and
```
susannasiebert, line 460: I believe the current logic will set `percentile_pass` to `False` if `self.percentile_threshold` is not set. But it should be set to `True` in that case.

pvactools/lib/aggregate_all_epitopes.py
```diff
@@ -452,6 +455,17 @@ def get_tier(self, mutation, vaf_clonal):
             if mutation["Transcript Support Level"] > self.maximum_transcript_support_level:
                 tsl_pass = False
 
+        ic50_pass = mutation["{} MT IC50 Score".format(self.mt_top_score_metric)] < binding_threshold
+        percentile_pass = (
+            self.percentile_threshold is not None and 
+            mutation["{} MT Percentile".format(self.mt_top_score_metric)] < self.percentile_threshold
+        )
+        thresholds_passed = (
```
susannasiebert, line 463: This is just for ease of reading the code but I would name this variable `binding_pass` since some of the other criteria (e.g. allele expr) also have thresholds/cutoffs.

pvactools/lib/filter.py
susannasiebert, line 1: This filter module is supposed to be generic so I think it is the wrong move to add details about the binding and percentile filters in here. I think it would be better to either (a) adjust merge the AlleleSpecificBindingFilter back into the BindingFilter and have it handle both the percentile strategy and the allele-specific and non-specific binding filtering or to (b) have the BindingFilter write its own logic about filtering similar to the AlleleSpecificBindingFilter. I'm leaning toward option (a) and have a BindingFilter that can handle all parameters.

pvactools/lib/aggregate_all_epitopes.py
```diff
@@ -452,6 +455,17 @@ def get_tier(self, mutation, vaf_clonal):
...
+            (ic50_pass and percentile_pass) 
+            if self.percentile_threshold_strategy == 'conservative' 
+            else (ic50_pass or percentile_pass)
+        )
+
```
susannasiebert, line 468: Might be getting too nitpicky here, but I think I would prefer to have this block of code be right after setting the `binding_threshold` (line 445) since they are logically related. I think it would make for easier readability.

pvactools/lib/filter.py
```diff
@@ -5,11 +5,12 @@
 pd.options.mode.chained_assignment = None
 
 class Filter:
-    def __init__(self, input_file, output_file, filter_criteria, int_filter_columns=[]):
+    def __init__(self, input_file, output_file, filter_criteria, int_filter_columns=[], filter_strategy=0):
```
susannasiebert, line 8: For easier code comprehension, I'm not a huge fan of using 0/1 here as the possible values for filter_strategy. I think using `AND`/`OR` or `any`/`all` would be more comprehensible.

ldhtnp:
> @ldhtnp I think this is ready to be merged in. What do you think?

@susannasiebert Yes, I think it is good to go!

malachig (approved):
Mostly reviewed the docs descriptions here.  It looks good to me.

susannasiebert (commented):
+1

susannasiebert:
Can you please make sure to document the new parameter in the [Aggregate Report Tiers](https://pvactools.readthedocs.io/en/latest/pvacseq/output_files.html#the-pvacseq-aggregate-report-tiers) section of the docs/pvacseq|fuse|splice|bind/output_files.rst files? It will need to be added in the Tiering Parameters section and the description of the "Binding Criteria" will need to be updated in the Criteria Details section.

Also, the new parameter and behavior should be described in the Binding Filter sections of the docs/pvacseq|fuse|splice|bind/filter_commands.rst files. 

It wouldn't hurt to grep the docs part of the repo for mention of the --percentile-threshold parameter to see if there are any other spots we describe this parameter that I forgot about.

PR #1186, opened 2025-01-30
Fix pVACvector logic for deleting tmp files

susannasiebert:
The previous logic was still referencing the `processed_spacers` variable which has been removed in previous versions. This update now fixes the logic and adds the removal to one more place in the code where it was previously missed. This PR also removes the `-k` parameter from one of the pVACvector tests which will result in this logic actually getting tested now.


ldhtnp (approved):
Looks good to me!

Issue #1187, opened 2025-01-31
Fix incorrect pVACview header tooltips


susannasiebert:
The tooltips currently have a off-by-one error due to the addition of the "Num Included" peptides. This should be fixed by adding a description of this new column. There is also a stray `Eval` description in there, which is a column that is no longer used. Other columns toward the right edge of the table are also missing since they were changed/added for the pVACview paper. The list should be compared to the actual currently available columns by loading the demo data into pvacview.org and adding columns that are missing, removing columns that no longer exist, and updating description text, where necessary.

PR #1189, opened 2025-01-31
Updated column tooltips

ldhtnp:
I added a tooltip for `Num Included Peptides` and replaced the old `Eval` tooltip with one for `Ref Match`.

Resolves issue [#1187](https://github.com/griffithlab/pVACtools/issues/1187)

### DIFF
--- a/pvactools/tools/pvacview/styling.R
+++ b/pvactools/tools/pvacview/styling.R
@@ -37,8 +37,8 @@ callback <- function(hla_count, score_mode) {
     "        'Rej - Click the thumbs-down button to reject a neoantigen candidate.',",
     "        'Rev - Click the flag button to mark a neoantigen candidate for review.'],",
     "header = table.columns().header();",
-    gsub("7", hla_count, "for (var i = 5; i-5 < tips.length; i++) {"),
-    gsub("7", hla_count, "$(header[i]).attr('title', tips[i-5]);"),
+    gsub("7", hla_count, "for (var i = 7; i-7 < tips.length; i++) {"),
+    gsub("7", hla_count, "$(header[i]).attr('title', tips[i-7]);"),
     "}"
   )
 }


pvactools/tools/pvacview/styling.R
```diff
@@ -20,6 +20,7 @@ callback <- function(hla_count, score_mode) {
     "        'Allele',",
     "        'Pos - The one-based position of the start of the mutation within the epitope sequence. 0 if the start of the mutation is before the epitope (as can occur downstream of frameshift mutations).',",
     "        'Prob Pos - Problematic positions within the best peptide.',",
+    "        'Num Included Peptides - The total number of peptides for this mutation.',",
```
susannasiebert, line 23: This might be a bit confusing for our users. While this does reflect the total number of peptides available in pVACview with meta-data, it's not all of the peptides resulting from the mutation.  In the context of the aggregated report this is described as:
> The number of included peptides according to the `--aggregate-inclusion-binding-threshold` and `--aggregate-inclusion-count-limit`

Here it might make sense to word it as "The number of top-scoring, unique peptides included in this report". @malachig @chrisamiller do you have opinions on this wording.

pvactools/tools/pvacview/styling.R
```diff
@@ -31,7 +32,7 @@ callback <- function(hla_count, score_mode) {
     "        'RNA Depth - Tumor RNA depth at this position.',",
     "        'DNA VAF - Tumor DNA variant allele frequency (VAF) at this position.',",
     "        'Tier - A tier suggesting the suitability of variants for use in vaccines.',",
-    "        'Eval - User-selected evaluation of neoantigen candidate. Options include: Accept, Reject, Review. (Default: Pending)'],",
+    "        'Ref Match - Indicates if the query sequence has a hit in the reference proteome.'],",
     "header = table.columns().header();",
     gsub("7", hla_count, "for (var i = 7; i-7 < tips.length; i++) {"),
     gsub("7", hla_count, "$(header[i]).attr('title', tips[i-7]);"),
```
susannasiebert, line 38: I'm not 100% sure whether the numbers here need adjusting. I think it has something to do with columns we drop for display purposes but I'm not 100% sure. If you haven't run pVACview with your modifications, please do to ensure that there are no errors or warnings related to this.

susannasiebert (changes requested):
I added a couple of change suggestions. I think it's also a good idea to add tooltips for the Acpt, Rej, and Rev columns. Something to effect of "Click the thumps-up button to accept a neoantigen candidate". etc.

Also, please change the base branch of the PR to the hotfix branch.

pvactools/tools/pvacview/styling.R
ldhtnp: The hla_count is now 5, so I updated the gsub commands to reflect this. I did notice when you click on a column value errors appear in the console. This occurred before and after I modified the gsub commands and also happens on the staging branch. Might be worth looking into

susannasiebert: I tried this change with a couple of different files and it looks like it now only works for files where the HLA count is the same as for the HCC1395 demo data. For anything with fewer or more HLAs the tooltips are shifted. If I change it back to how it was originally, it seems to work fine aside from the messages you noticed in the console. I think those are unrelated.

PR #1198, opened 2025-02-07
include non-human transcripts to pvacsplice

mhoang22:
previously combine_input.py searched for transcript_id starts with 'ENST' in the annotated_vcf(tsv). this would exclude non-human transcripts.
we change the code so that it searches for transcript_id contains ENS*T instead, to include non-human transcripts.

Notes on ensembl transcript name:
https://useast.ensembl.org/info/genome/stable_ids/prefixes.html 
https://useast.ensembl.org/Help/Faq?id=488

(example: mice: ENSMUST, dog: ENSCAFT, cat: ENSFCAT)

### DIFF
--- a/pvactools/lib/splice_pipeline.py
+++ b/pvactools/lib/splice_pipeline.py
@@ -1,3 +1,4 @@
+import sys
 import shutil
 import os
 import pandas as pd


susannasiebert (commented):
+1 this looks good but I wonder if we should add some non-human tests here to cover this specific case. In tests/test_pvacsplice_combine_outputs.py you can basically replicate/copy the `test_combine_inputs_runs_and_produces_expected_output` test and use your dog test files if we're ok with putting them publicly on GitHub. You'll need to use intermediate files produced by pVACsplice for the junctions and variant file inputs.

susannasiebert:
I think this is also a good PR to fix the missing `import sys` in `lib/splice_pipeline.py`.

susannasiebert (approved):
+1. Non-human tests will be added as part of #1207

Issue #1194, opened 2025-02-04
integrate pVACcompare to pVACtools


mhoang22:
review and add more features for ease of use

malachig:
To test with different versions of the software there are many example results here:
/storage1/fs1/gillandersw/Active/Project_0001_Clinical_Trials/pipeline_tests/hcc1395/pvactools_releases

In addition we should test with some multi-region tumor cases (e.g. GBMs):
/storage1/fs1/gpdunn/Active/Project_0001_Clinical/gc2609/gbm_antigen_dunn: vaccine_G103, vaccine_G104, vaccine_G110, vaccine_G113

PR #1203, opened 2025-02-17
Integrated pVACcompare into pVACtools

ldhtnp:
Issue [#1194](https://github.com/griffithlab/pVACtools/issues/1194)


pvactools/tools/compare.py
```diff
@@ -0,0 +1,231 @@
...
+        "Tumor RNA VAF",
+        "Gene Expression",
+    ]
+    default_reference_match_columns = ["Peptide", "Match Window"]
+
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.add_argument(
+        "results_folder1", nargs="?", help="Path to first results input folder"
```
susannasiebert, line 127: What's the rational for using `nargs="?"` here and for results_folder2?

pvactools/tools/pvaccompare/server.py
```diff
@@ -0,0 +1,38 @@
...
+        self.send_response(200)
+        self.end_headers()
+
+
+def main():
+    PORT = 8080
+    server = HTTPServer(("localhost", PORT), CORSRequestHandler)
+    print("Starting local server...")
+    print(
+        f"View reports at http://localhost:{PORT}/pvactools/tools/pvaccompare/html_report/main.html"
```
susannasiebert, line 27: This will only work if you happen to start the server from a directory that contains this file path underneath it, which wont' be the case once pVACtools is installed somewhere. You will need to get the actual absolute installation path of this file on the user's system or copy the main.html from its installation path to the working directory before starting the server (and then clean it up at KeyboardInterrupt).

pvactools/tools/compare.py
```diff
@@ -0,0 +1,231 @@
...
+        "Best Transcript",
+        "Num Passing Peptides",
+        "IC50 MT",
+        "IC50 WT",
+        "%%ile MT",
+        "%%ile WT",
+        "RNA Expr",
+        "RNA VAF",
+        "DNA VAF",
+        "Tier",
```
susannasiebert, line 32: We might want to add the `Ref Match` column here (`True`/`False`).

pvactools/tools/compare.py
```diff
@@ -0,0 +1,231 @@
...
+        "SMM MT IC50 Score",
+        "SMM WT Percentile",
+        "SMM MT Percentile",
+        "SMMPMBEC WT IC50 Score",
+        "SMMPMBEC MT IC50 Score",
+        "SMMPMBEC WT Percentile",
+        "SMMPMBEC MT Percentile",
+        "DeepImmuno WT Score",
+        "DeepImmuno MT Score",
+        "Problematic Positions",
```
susannasiebert, line 93: Version 5.4 introduces a new features that would add a `Gene of Interest` column here that might be interesting to compare. The possible values are `True`/`False`.

PR #1228, opened 2025-04-04
Fix invalid literal for int() with base 10 error

riederd:
This fixes an error in `aggregate_all_epitopes.py` when the `Mutation Position` is stored as floating point number in `_all_epitopes.tsv` (see attached example)
[HCC1395_MHC_Class_II_all_epitopes_ex.tsv.gz](https://github.com/user-attachments/files/19600635/HCC1395_MHC_Class_II_all_epitopes_ex.tsv.gz)

error:

```
  File "/opt/conda/lib/python3.11/site-packages/pvactools/lib/aggregate_all_epitopes.py", line 429, in <genexpr>
    if all(int(pos) in anchors for pos in positions):
           ^^^^^^^^
ValueError: invalid literal for int() with base 10: '15.0'
````

### DIFF
--- a/pvactools/lib/aggregate_all_epitopes.py
+++ b/pvactools/lib/aggregate_all_epitopes.py
@@ -426,7 +426,7 @@ class PvacseqAggregateAllEpitopes(AggregateAllEpitopes, metaclass=ABCMeta):
             if len(positions) > 2:
                 return True
             anchor_residue_pass = True
-            if all(int(pos) in anchors for pos in positions):
+            if all(int(float(pos)) in anchors for pos in positions):
                 if pd.isna(mutation["{} WT IC50 Score".format(self.wt_top_score_metric)]):
                     anchor_residue_pass = False
                 elif mutation["{} WT IC50 Score".format(self.wt_top_score_metric)] < binding_threshold:


susannasiebert:
@riederd thank you for this pull request. Out of curiosity, is this an error that you encountered in the full `pvacseq run` pipeline? Or did you by chance open this file in Excel or similar and this column got converted to a float on save/export? These numbers shouldn't be stored as a decimal so I would like to fix that if that is happening in the full pipeline.

riederd:
Oh, good call !
I created input file for `pvacseq  generate_aggregated_report` by concatenating the output TSVs from multiple pVACseq runs using a python/pandas script. I guess pandas converted to decimals.

susannasiebert:
Ah yes, that would explain it. Pandas tries to automatically detect the data type but its notorious for not getting it quite right. For this field, you want to treat it as a string because for non-missense mutation this column could contain a comma-separated list of positions.

susannasiebert:
@riederd were you able to concatenate your files in a way that works with the existing `pvacseq generate_aggregated_report` command?

riederd:
Yes it works if I use explicitly tell pandas to read the pVACseq result tsv files as string type (`pd.read_csv(file, sep="\t", dtype=str)`)

Thanks

Issue #1193, opened 2025-02-04
package the peptide order form assist scripts (create report and color peptide) into a tool


mhoang22:

malachig:
Specifically talking about these steps currently done in a more ad hoc fasion:

```
bsub -Is -q oncology-interactive -G $GROUP -a "docker(griffithlab/neoang_scripts)" /bin/bash

cd $WORKING_BASE
mkdir ../manual_review

python3 /opt/scripts/generate_reviews_files.py -a ../itb-review-files/*.xlsx -c ../generate_protein_fasta/candidates/annotated_filtered.vcf-pass-51mer.fa.manufacturability.tsv -variants final_results/variants.final.annotated.tsv -classI final_results/pVACseq/mhc_i/*.all_epitopes.aggregated.tsv -classII final_results/pVACseq/mhc_ii/*.all_epitopes.aggregated.tsv -samp $PATIENT_ID -o ../manual_review/

python3 /opt/scripts/color_peptides51mer.py -p ../manual_review/*Peptides_51-mer.xlsx -samp $PATIENT_ID -o ../manual_review/
```

susannasiebert:
The repo with the scripts can be found here: https://github.com/griffithlab/neoag_vaccine_scripts

susannasiebert:
Writing up some more detailed implementation notes:
After the Immunotherapy Tumor Board meeting, the peptide order from is created in a three-step process:
1) Create a peptide fasta file with a larger number of flanking amino acids (typically 25) using the `pvacseq generate_protein_fasta` command
2) Generate a peptide ordering Excel sheet using the neoag_vaccine_scripts `generate_reviews_files.py` script
3) Color the peptide ordering Excel sheet using the neoag_vaccine_scripts `color_peptides51mer.py` script

The goal of this PR is to move this process to a single command in the pVACtools package. Since this process currently only applies to pVACseq results, I think this command should live in the `pvacseq` command line tree. A possible name for the new command would be `pvacseq create_peptide_ordering_form`.

This new command would need to first call `pvacseq generate_protein_fasta`. I don't think we want to just shell out here but pvactools/tools/pvacseq/generate_protein_fasta.py isn't actually a class so it will probably require some modifications to get that to work. Similar issue for the neoag_vaccine_scripts.

In terms of inputs, the generate_protein_fasta steps requires the following inputs:
- A annotated VCF file (the same file used to start `pvacseq run`
[...]

PR #1245, opened 2025-05-30
Added create_peptide_ordering_form command to pVACseq

ldhtnp:
Added the create_peptide_ordering_form command to pVACseq, which packages the peptide order form assist scripts into a single tool.

Mentioned in issue [#1193](https://github.com/griffithlab/pVACtools/issues/1193)


pvactools/lib/color_peptides51mer.py
```diff
@@ -0,0 +1,386 @@
...
+        
+        if len(probPos) > 0 and sequence[i] in probPos:
+            new_AA.large = True
+        
+        peptide_sequence.append(new_AA)
+
+    # CLASS I
+    positions = []
+    # Get the positions in the peptide_sequence where the classI is located
+    for i in range(len(peptide_sequence)):
```
susannasiebert, line 48: I think this could be simplified to something like:
```
sequence = [s.nucleotide for s in peptide_sequence]
start = sequence.index(classI_peptide)
positions = list(range(start, start + len(b)))
```

pvactools/lib/color_peptides51mer.py
```diff
@@ -0,0 +1,386 @@
...
+            positions = []
+            
+    # set those positions to red
+    # if median affinity < 1000 nm OR percentile < 2%
+    j = 0
+    if float(classI_ic50) < cIIC50_threshold or float(classI_percentile) < cIpercentile_threshold:
+        for i in range(len(peptide_sequence)):
+            if j < len(positions) and i == positions[j]:
+                peptide_sequence[i].color = True
+                j+=1
```
susannasiebert, line 68: Can this be?:
```
for position in positions:
  peptide_sequence[position].color = True
````

File: pvactools/lib/color_peptides51mer.py [hunk 3]
```diff
@@ -0,0 +1,386 @@
...
+            else:
+                positions = []
+        # Set class II to bold
+        # if percentile < 2% 
+        j = 0
+        if float(classII_percentile) < cIIpercent_threshold or float(classII_ic50) < cIIIC50_threshold:
+            for i in range(len(peptide_sequence)):
+                if j < len(positions) and i == positions[j]:
+                    peptide_sequence[i].bold = True
+                    j+=1
```
susannasiebert, line 92: Same as above.

pvactools/lib/color_peptides51mer.py
```diff
@@ -0,0 +1,386 @@
...
+
+    frameshift = False
+    classI_position = 0
+
+    # Determine if frameshift mutation by seraching for = '-'
+    if ',' in mutant_peptide_pos:
+        positions = mutant_peptide_pos.split(",")
+        start_position = int(positions[0])
+        end_position = int(positions[1])
+        frameshift = True
```
susannasiebert, line 110: This seems incorrect to me and should be further debugged:
- I think the separator needs to be `, `.
- The mutant peptide position column can have more than two values so the second value is not necessarily the end position. It should be `positions[-1]`.
- Having more than two mutated positions does not necessarily mean that this is a frameshift. It could be a larger indel or a missense mutation with a proximal variant. It could even be a frameshift with a proximal variant in the upstream sequence (before the framshift). There are other fields that contaim information about the variant type that would be better suited to setting frameshift True/False.
- Because a frameshift could have an upstream proximal variant, we need to be careful about the underlining done in lines 121+. We don't want to include the proximal variant's amino acid(s), and any unmutated amino acids in between in the underlining. So I think the logic needs to start with the last mutated position, add all consecutive mutated positions in before of it and all positions after it.

I suggest carefully investigating some inframe indels and frameshifts with some of the caveats described above and checking that it does the right thing.

susannasiebert (changes requested):
I did some preliminary review and noted some issues and possible improvements I found.

I would also carefully investigate why the test excel sheet doesn't have some of the entries colored, bolded, underlined etc. Are these supposed to be missing those annotations because they're specific edge cases we are covering or are they bugs in the logic? Same with the NA values for the PRICKLE4 variant. If they are testing edge case handling, I think it would be preferred to isolate them and test them separately.

Issue #1221, opened 2025-03-13
Output files should include mhc_i/mhc_ii/combined


susannasiebert:
When running with both MHC Class I and Class II alleles/algorithms, the output file names are indistinguishable outside of their parent directory names. We should include mhc_i/mhc_ii/combined in the main report file names to make these files easier to identify once they are moved into different locations.

PR #1252, opened 2025-06-19
Main report file names given MHC_I/MHC_II/Combined Prefix (Issue #1221)

Jaz2021:
Added a parameter to the pipeline so that you can add a given prefix to the output. Currently used for MHC_I, MHC_II, and Combined output files, but can be easily updated for any future use cases.


susannasiebert (changes requested):
Overall this looks good. The same change should also be made for the other pipeleines. Please ask Malachi about his preference on the order of the sample name vs MHC class in the file name when he gets back from vacation. You will also want to edit the documentation, specifically the output_files.rst section of each of the pipelines to update the file names.

docs/pvacbind/output_files.rst
```diff
@@ -20,16 +20,16 @@ created):
 
    * - File Name
      - Description
-   * - ``<sample_name>.all_epitopes.tsv``
+   * - ``<sample_name>.<MHC_Class/Combined>.all_epitopes.tsv``
```
susannasiebert, line 23:
```suggestion
   * - ``<sample_name>.<MHC_I|MHC_II|Combined>.all_epitopes.tsv``
```
I think this wording is a bit clearer. If you agree, please update throughout the docs.

susannasiebert (approved):
One small suggestion for the docs but otherwise it looks good.

Issue #1257, opened 2025-06-25
Sorting on the IC50/%ile WT columns should be numerical, not string sorting.


susannasiebert:
Not sure if that is possible when there are NA values but it should be investigated.

PR #1265, opened 2025-07-07
Fixed IC50/%ile WT sorting

ldhtnp:
Fixed sorting of the IC50/%ile WT columns when NA values are present

Resolves issue ([#1257](https://github.com/griffithlab/pVACtools/issues/1257))


susannasiebert (approved):
+1

Issue #1264, opened 2025-07-03
Filter out transcript with incomplete CDS 5' and/or 3'


susannasiebert:
Some transcript are annotated with an incomplete CDS 5' or 3' position. This is reflected, for example, by the first amino acid being X in the reference peptide fasta. Such transcripts are not desired and should be filtered out.

VEP annotates such transcripts in the FLAGS  (transcript quality) CSQ field. This can be either empty or have values of `cds_start_NF` and/or `cds_end_NF`. If these values are set, transcript should not be included for processing in pVACseq.

TBD, the exact behavior here. Some options are:
- hard-filter these upstream when we parse the VCF
- hard-filter these upstream but only when a new variable is set
- add a new option to the transcript prioritization strategy (transcript_quality, maybe) that will deprioritize transcripts with a FLAGS set

malachig:
I think it would be great to have option (2) with a new parameter called: `--allow-incomplete-transcripts`.  

The default value would be FALSE and these transcripts would be filtered out.  BUT, if the user sets it to TRUE, peptides from these transcripts are allowed (not containing and `X` but from other parts of the predicted protein sequence).

If `--allow-incomplete-transcripts` is true, candidates coming from these transcripts should be deprioritized compared to transcripts that do NOT have these FLAGs defined.

For documentation we want to explain that this option refers to the Ensembl transcript annotations "CDS 5' and/or 3' incomplete" (aka "cds_start_NF" or "cds_end_NF".

susannasiebert:
Extra points if this behavior also gets implemented for pVACsplice. My ran into the same issue with some pVACsplice runs. You might be able to use the same test files I created for https://github.com/griffithlab/pVACtools/pull/1268 that cover this issue.

PR #1293, opened 2025-08-25
Added new parameter to handle filtering of incomplete CDS transcripts

ldhtnp:
Added the parameter `--allow-incomplete-transcripts` to both pVACseq and pVACsplice. By default when this parameter is not specified transcripts with incomplete CDS flags will be filtered out. When the parameter is specified these transcripts will be included but deprioritized compared to transcripts that do not have these flags defined.

Mentioned in issue #1264


pvactools/lib/load_gtf_data.py
```diff
@@ -37,6 +38,13 @@ def execute(self):
                 (gtf_df_all['feature'].isin(['CDS', 'transcript']))
                 ].replace(["^\s*$"], 'NA', regex=True)
             gtf_df['transcript_biotype'] = 'Not Supported'
+        
+        if 'tag' in gtf_df_all:
+            if self.allow_incomplete_transcripts:
+                gtf_df.loc[~gtf_df['tag'].str.lower().str.contains("cds_start_nf|cds_end_nf", na=False), 'tag'] = "None"
+            else:
+                gtf_df = gtf_df.loc[~gtf_df['tag'].str.lower().str.contains("cds_start_nf|cds_end_nf", na=False)].copy()
+                gtf_df['tag'] = "NA"
```
susannasiebert, line 48: This logic doesn't seem to match the logic for the VCF. We also need to decide what to do when the `tag` field is missing.

susannasiebert (approved):
+1

PR #1295, opened 2025-08-26
Improved compare input file search

ldhtnp:
Reworked the input file search logic to recursively search for the class files instead of looking at exact file paths.


susannasiebert (approved):
+1

PR #1299, opened 2025-08-29
Fixed input yml parsing

ldhtnp:
Modified the script to be able to locate the input yml file after a pipeline run


susannasiebert (approved):
Made one small suggestion but otherwise it looks good. +1 assuming tests pass.

PR #1308, opened 2025-09-16
Modified FS underlining, fixed deletion start indexing, removed space before leading format

ldhtnp:
Modified the underlining logic for frameshifts to underline from the first mutated position to the end of the sequence. Modified the  `affected_start` for deletions to match pVACtools. Changed the space inserted before a leading format to the unicode zero-width space `\u200b`.


susannasiebert (approved):
+1

PR #1333, opened 2025-11-03
ML predictor integration into pvacseq

jyao36:
…odel artifacts; modified argument parser and pvacseq run.py file to run the predictor as part of pvacseq; added test files and script.


susannasiebert (commented):
This looks pretty good. I added some comments for improvements that could be made.

I would also like to see the addition of a standalone `pvacseq add_ml_predictions` command that someone could run standalone to add ml_predictions to the output from their pipeline if the ml option was not enabled in the original run. Have a look at `pvactools/tools/pvacseq/mark_genes_of_interest.py ` and its test `tests/test_pvacseq_mark_genes_of_interest.py` to see how to do so. This command addition also requires an update to `pvactools/tools/pvacseq/main.py` and `pvactools/tools/pvacseq/__init__.py`  to add this command.

Additionally, this new command should then be documented in `docs/pvacseq/optional_downstream_analysis_tools.rst`. Most of the documentation can be auto-generated from the command line docs by using the program-output tool but this would also be the spot to put any additional documentation you want to have around this feature.

pvactools/lib/ml_predictor.py
```diff
@@ -0,0 +1,447 @@
...
+    parser.add_argument(
+        "--class1-aggregated",
+        required=True,
+        dest="class1_aggregated",
+        help="Path to the MHC Class I aggregated epitopes TSV."
+    )
+    parser.add_argument(
+        "--class1-all-epitopes",
+        required=True,
+        dest="class1_all_epitopes",
```
susannasiebert:
```suggestion
        "class1_all_epitopes",
```

pvactools/lib/ml_predictor.py
```diff
@@ -0,0 +1,447 @@
...
+        help="Path to the MHC Class I all epitopes TSV."
+    )
+    parser.add_argument(
+        "--class2-aggregated",
+        required=True,
+        dest="class2_aggregated",
```
susannasiebert:
```suggestion
        "class2_aggregated",
```

susannasiebert (commented):
Left a couple of code suggestions to convert named required parameters to positional parameters in your parser.

tests/test_data/ml_predictor/ml_predictor_test_output/HCC1395_predict_pvacview.tsv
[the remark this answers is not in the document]
susannasiebert: This looks a lot better. The only issue I see is that  Prob Pos is now `NA` but it should be `None` when there are no problematic positions.

susannasiebert (approved):
+1

docs/pvacseq/optional_downstream_analysis_tools.rst
```diff
@@ -61,6 +61,39 @@ epitopes are well-binding to. Lastly, the report will bin variants into tiers
...
+"Reject", or "Pending" based on a comprehensive set of features derived from binding affinity predictions, 
+expression data, and variant characteristics.
+
+This tool requires that you have already generated both MHC Class I and Class II aggregated reports using 
+the ``generate_aggregated_report`` command. It takes as input the Class I aggregated TSV, Class I all epitopes TSV, 
+and Class II aggregated TSV files from a pVACseq run. The tool merges these files, performs data cleaning and 
+imputation, and applies the ML model to generate evaluation predictions for each variant.
+
+The output file is named ``<sample_name>_predict_pvacview.tsv`` and contains all columns from the original 
+Class I aggregated file with two differentcolumns:
```
susannasiebert:
```suggestion
Class I aggregated file with two additional columns:
```

docs/pvacseq/output_files.rst
```diff
@@ -387,6 +390,25 @@ included epitopes, selecting the best-scoring epitope, and which values are outp
    * - ``Evaluation``
      - Column to store the evaluation of each variant when evaluating the run in pVACview. Either ``Accept``, ``Reject``, or ``Review``.
 
+.. _ml_prediction_output:
+
+<sample_name>_predict_pvacview.tsv Report Columns
+--------------------------------------------------
+
+The ``<sample_name>_predict_pvacview.tsv`` file is generated when using the :ref:`add_ml_predictions <optional_downstream_analysis_tools_label>` tool or when running pVACseq with both MHC Class I and Class II predictions and the ``--run-ml-predictions`` flag enabled. This file contains all columns from the Class I aggregated file (``all_epitopes.aggregated.tsv``) with one additional ML prediction columns added.
```
susannasiebert:
```suggestion
The ``<sample_name>_predict_pvacview.tsv`` file is generated when using the :ref:`add_ml_predictions <optional_downstream_analysis_tools_label>` tool or when running pVACseq with both MHC Class I and Class II predictions and the ``--run-ml-predictions`` flag enabled. This file contains all columns from the Class I aggregated file (``all_epitopes.aggregated.tsv``) with one additional ML prediction column added.
```

pvactools/lib/run_argument_parser.py
```diff
@@ -380,6 +380,18 @@ def pvacseq(self):
...
+            "--run-ml-predictions",
+            help="Enable ML-based neoantigen evaluation predictions.",
+            default=False,
+            action='store_true',
+        )
+        self.parser.add_argument(
+            "--ml-threshold", type=float,
+            default=0.55,
+            help="Threshold for Accept predictions in ML model (default: 0.55).",
+        )
```
susannasiebert: I wonder if we should consider making both the Accept and the Reject threshold configurable. Right now the reject threshold is hardcoded to 0.3.

susannasiebert (changes requested):
I made some copy editing suggestions. Two thoughts I had while reviewing this:
- It would be great to have some additional exploration of different candidates in ML section of the pVACview vignette. I wrote down some examples in a comment in that part of the text.
- Have you and Malachi discussed whether it would be worthwhile to make the reject threshold configurable? It seems a bit weird to me to hardcode the reject threshold but make the accept threshold configurable.

pvactools/lib/run_argument_parser.py
jyao36: Will incorporate that, don't think there's any harm doing that.

susannasiebert (approved):
+1

docs/pvacseq/output_files.rst
```diff
@@ -55,6 +55,9 @@ created):
    * - ``ui.R``, ``app.R``, ``server.R``, ``styling.R``, ``anchor_and_helper_functions.R``
      - pVACview R Shiny application files. Not generated when running only with presentation and immunogenicity algorithms.
    * - ``www`` (directory)
+     - Directory containing image files for pVACview. Not generated when running with presentation and immunogenicity algorithms only.  
+   * - ``ml_predict/<sample_name>_predict_pvacview.tsv`` (optional)
+     - ML-based neoantigen evaluation predictions file. Generated when both MHC Class I and Class II predictions are run and the ``--run-ml-predictions`` flag is set.
      - Directory containing image files for pVACview. Not generated when running only with presentation and immunogenicity algorithms only.
```
jyao36, line 60: This table doesn't seem to be showing up on the webpage

docs/pvacseq/output_files.rst
```diff
@@ -55,6 +55,9 @@ created):
...
```
susannasiebert:
```suggestion
```

This should fix the issue with this table not rendering.

docs/pvacseq/output_files.rst
```diff
@@ -55,7 +54,9 @@ created):
    * - ``ui.R``, ``app.R``, ``server.R``, ``styling.R``, ``anchor_and_helper_functions.R``
      - pVACview R Shiny application files. Not generated when running only with presentation and immunogenicity algorithms.
    * - ``www`` (directory)
-     - Directory containing image files for pVACview. Not generated when running only with presentation and immunogenicity algorithms only.
+     - Directory containing image files for pVACview. Not generated when running with presentation and immunogenicity algorithms only.  
+   * - ``MHC_Class_I/<sample_name>.MHC_I.all_epitopes.aggregated.ML_predict.tsv`` (optional)
+     - A version of the ``<sample_name>.MHC_I.all_epitopes.aggregated.tsv`` with ML-based neoantigen evaluation predictions. Generated when both MHC Class I and Class II predictions are run and the ``--run-ml-predictions`` flag is set. Written to the same folder as the Class I aggregated file.
```
susannasiebert:
```suggestion
   * - ``<sample_name>.MHC_I.all_epitopes.aggregated.ML_predict.tsv`` (optional)
     - A version of the ``<sample_name>.MHC_I.all_epitopes.aggregated.tsv`` with ML-based neoantigen evaluation predictions. Generated when both MHC Class I and Class II predictions are run and the ``--run-ml-predictions`` flag is set. Written only to the ``MHC_Class_I`` folder.
```

pvactools/lib/ml_predictor.py
```diff
@@ -477,7 +477,9 @@ def define_add_ml_predictions_parser(tool='pvacseq'):
     )
     parser.add_argument(
         "output_dir",
-        help="Directory where the ML prediction TSV files should be written."
+        nargs="?",
```
susannasiebert: Instead of using `nargs="?"` and a positional parameter for optional parameters, we've been using the `--` convention (like for the `--ml-threshold-accept` and `--ml-threshold-predict`). If you rename the parameter to `--output-dir` it will automatically be optional. It should also be moved behind all of the positional parameters, i.e between `sample_name` and `--ml-threshold-accept`).

susannasiebert (changes requested):
This looks good. Just two minor change requests.

susannasiebert (approved):
+1

PR #1341, opened 2025-11-17
Use Index to construct 51mer ID instead of modifying AA change

susannasiebert:
By using the Index, we can use the existing `modify_id` method for converting to our desired format and ensure that the end result (51mer ID) in the class I aggregated report (`classI`) df is the same as the 51mer ID created in the peptides TSV (`peptides`) df. 

I believe the AA change was used originally because the Index was added to the aggregated report more recently.


susannasiebert:
Ok, looking into this. I noticed that the colored sheet has some variants listed twice (e.g. MT.931.IL27RA.ENST00000263379.4.missense.428A/V and MT.932.IL27RA.ENST00000263379.4.missense.428A/V are each listed twice; they're part of the DNV and result from two different variant lines in the VCF). I think the problem is that the 51mer ID will be identical for these two entries so when combining entries using the 51mer ID it is doubling them up. I think fixing this will also fix the issue you are seeing.

PR #1351, opened 2025-12-16
Implement an alternate way to calculate normalized percentile scores

susannasiebert:
In the first implementation, the scores were being calculate on a per-length basis. This update calculates them in a length-agnostic way. The update was implemented via new parameter to the `calculate_normalized_percentile` method. This will allow us to run the normalized percentile calculation either way. The parameter can be exposed to our users in the run commands, if so desired.

It's still to be determined, which way is preferable.


susannasiebert:
It's been decided to move forward with this way to calculate normalized percentiles.

ldhtnp (approved):
+1
Changes look good to me!

PR #1371, opened 2026-03-07
Added the ImmuScope class II prediction algorithm

### LINKED ISSUE #1330: Investigate ImmuScope for class II predictions
https://github.com/shenlongchen/ImmuScope

ldhtnp:
I created a [fork](https://github.com/griffithlab/ImmuScope) of the [original](https://github.com/shenlongchen/ImmuScope) ImmuScope repository to refine it for use in pVACtools. This involved creating a wrapper to allow ImmuScope to output an immunogenicity score given a peptide + hla pair. I based the implementation off of BigMHC_IM and updated the documentation.

This tool was suggested by Malachi in issue #1330


docs/install.rst
```diff
@@ -232,6 +232,37 @@ You can check that DeepImmuno was installed successfully by running:
 
 This should show information about the DeepImmuno installation.
 
+Installing Immuscope
+---------------------
+
+If you wish to run the Immuscope_IM prediction algorithm, you will need to
+install Immuscope on your system. This package not a direct dependency of
```
susannasiebert:
```suggestion
install Immuscope on your system. This package is not a direct dependency of
```

predictor_tests/test_call_iedb.py
```diff
@@ -397,5 +397,35 @@ def test_netmhciipan_el_method_with_version(self):
...
+            self.input_file,
+            call_iedb_output_file.name,
+            'ImmuScope_IM',
+            'DRB1*01:01',
+            '-l', '15',
+        ])
+        expected_output_file = os.path.join(self.test_data_dir, 'output_immuscope_im.tsv')
+        expected_df = pd.read_csv(expected_output_file, sep="\t", index_col=[1,5,6])
+        actual_df = pd.read_csv(call_iedb_output_file.name, sep="\t", index_col=[1,5,6])
+        pd.testing.assert_frame_equal(expected_df, actual_df, check_like=True, check_exact=False)
```
susannasiebert, line 426: These two test cases seem to be identical.

pvactools/lib/prediction_class.py
```diff
@@ -822,6 +822,128 @@ def iedb_executable_params(self, iedb_executable_path, method, allele, input_fil
...
+            all_epitopes.extend(epitopes.values())
+        all_epitopes = list(set(all_epitopes))
+
+        if len(all_epitopes) == 0:
+            return (results, 'pandas')
+
+        tmp_input_file = tempfile.NamedTemporaryFile('w', dir=tmp_dir, delete=False)
+        tmp_input_file.write("allele\tpeptide\tseq_num\tstart\n")
+        for epitope in all_epitopes:
+            tmp_input_file.write("{}\t{}\t\t\n".format(allele, epitope))
```
susannasiebert: I would prefer to use a csv.writer here.

susannasiebert (changes requested):
I added a few small suggested changes to the code itself.

I would like to see the output from this tool as an input file to the all class ii output parser test. To create this input file you will need to run ImmuScope on the 1-200 fasta chunk of HCC1395 data using allele DRB1*04:05 and length 12. I can send you this file.

The all class ii output file created by this test should then be used as the updated input to the  all class ii aggregate report creation test. Updating these two tests will ensure that ImmuScope gets parsed correctly and its detailed data included in the metrics file. Edit to add: I just realized that this test is not available in your branch. I added it as part of #1376 so you can disregard this comment.

Have you run HCC1395 on this PR? If not, I can make a test docker container once these updates have been made and start a run. I like to load to results into pVACview to check that everything looks as expected.

pvactools/lib/prediction_class.py
```diff
@@ -884,10 +884,11 @@ def predict(self, input_file, allele, epitope_length, iedb_executable_path, iedb
         if len(all_epitopes) == 0:
             return (results, 'pandas')
 
-        tmp_input_file = tempfile.NamedTemporaryFile('w', dir=tmp_dir, delete=False)
-        tmp_input_file.write("allele\tpeptide\tseq_num\tstart\n")
+        tmp_input_file = tempfile.NamedTemporaryFile('w', dir=tmp_dir, delete=False, newline='')
+        writer = csv.writer(tmp_input_file, delimiter='\t', lineterminator='\n')
+        writer.writerow(["allele", "peptide", "seq_num", "start"])
```
susannasiebert, line 899: Is it required by the predictor to add the seq_num and start to the input file? If not I think these columns can be removed.

ldhtnp: Yes, they are currently required by the predictor/wrapper interface as implemented. If you would rather them be excluded, I can update the fork to make these optional

susannasiebert: Gotcha. I think we could generate the file with these columns filled in by creating it in the same block of code above where we read in the fasta file (line 878+). The `determine_neoepitopes` method returns a hash with the start position as the key and the epitope as the value. The fasta sequence header can be used as the seq_num. 

I assume that the output includes these two columns as well so that would then save us from having to map back each epitope to it's seq num and start position (line 934+). This would be at the expense of potentially having duplicate epitopes in that file if there are repetitive regions etc which could make ImmunoScope slower (not sure if they accounted for this).

ldhtnp: I pushed a commit that keeps the deduped peptide set for scoring, but captures seq_num and start during the initial FASTA parsing and then merges them back onto the ImmuScope output. This lets us drop the remapping loop while still preserving those fields cleanly.

The performance of ImmuScope would be impacted if we passed every epitope occurrence directly to the wrapper with seq_num/start filled in, since it would score duplicates instead of just unique peptides. This approach avoids that by keeping the input deduplicated and only expanding back afterward.

susannasiebert: I wasn't sure if Immuscope was being smart and deduplicates epitopes on their end.

susannasiebert (approved):
+1

PR #1378, opened 2026-03-17
Remove pdb debugger from output_parser.py

maksimf:
## Summary
- Removes `pdb.set_trace()` debug code left in production in `match_wildtype_and_mutant_entry_for_missense()`
- Replaces the bare `except: pdb.set_trace()` with a proper `KeyError` that includes the missing position and MT epitope sequence for easier debugging

## Problem
Any exception during WT result lookup would freeze execution by dropping into an interactive debugger (`pdb`), making the tool hang indefinitely in production/CI environments.

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -28,3 +28,4 @@ Release Notes
    releases/5_4
    releases/5_5
    releases/6_0
+   releases/6_1
--- a/pvactools/lib/output_parser.py
+++ b/pvactools/lib/output_parser.py
@@ -210,9 +210,8 @@ class OutputParser(metaclass=ABCMeta):
         mt_epitope_seq = result['mt_epitope_seq']
         try:
             wt_result      = wt_results[match_position]
-        except:
-            import pdb
-            pdb.set_trace()
+        except KeyError:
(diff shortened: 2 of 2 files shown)


susannasiebert:
@maksimf Thank you for this pull request.

This breakpoint should really never be hit as it points to a fundamental mismatch between the output files from the chosen predictors and the fasta file generated by pVACseq that is used as input to those predictor calls which shouldn't happen during a normal pVACseq run (from scratch). There are two circumstance where I can imagine hitting this breakpoint/error:
- pVACseq was rerun with different pre-filtering parameters (e.g. --pass-only, --biotypes, --allow-incomplete-transcripts) on an existing run with existing predictor output files.
- pVACseq was rerun with a different version on an existing run with existing predictor output files

In both situations pVACseq should be run from scratch.

Could you please provide me some additional information for the circumstance under which you encountered this issue (hanging run in production)?

maksimf:
Thank you for the reply, Susanna. To be completely transparent I did not encounter this bug myself, I read through the code and the breakpoint caught my eye, as I thought it was left there unintentionally. If I was wrong then by all means let's close this PR

susannasiebert:
I think changing it to an actual error is nicer so I have no objections to this PR.

susannasiebert (approved):
+1

PR #1380, opened 2026-03-18
Enforce pVACvector pathfinder preconditions

maksimf:
## Summary
- Add a fail-fast precondition check in `find_optimal_path` to ensure it is only called with a graph that passes `check_graph_valid`.
- Raise a clear exception when `find_optimal_path` is invoked with an invalid graph state.
- Update the regression test to validate the expected exception for this out-of-contract invocation.

### DIFF
--- a/pvactools/tools/pvacvector/run.py
+++ b/pvactools/tools/pvacvector/run.py
@@ -257,6 +257,10 @@ def create_distance_matrix(Paths):
     return distance_matrix
 
 def find_optimal_path(graph, distance_matrix, seq_dict, base_output_dir, junctions_file, args):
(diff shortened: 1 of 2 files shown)


susannasiebert:
`find_optimal_path` should only be called if `check_graph_valid` returns True for its first return value, as you can see in `main` method. If anything the case that you are trying to test here should result in a fatal error at the beginning of `find_optimal_path` as this method should only be called if the graph contains nodes for all desired peptides. So something like:
```
if sorted(list(graph.nodes())) != sorted(list(seq_dict.keys()))
  raise Exception("Graph doesn't contain nodes for all peptides")
```

maksimf:
@susannasiebert Thank you Susanna, is this better?

PR #1384, opened 2026-03-27
Additional pVACview improvements

susannasiebert:


susannasiebert:
Note to self: prevent clicking on transcript table rows.

PR #1402, opened 2026-05-29
Resolved nondeterministic sorting in reference match comparisons

ldhtnp:
This PR fixes a flaky test failure in the reference match comparison tests that was showing up in the Python 3.9 GitHub Actions workflow.

The issue came from nondeterministic ordering when multiple records shared the same chromosome/start/stop values. Sorting now uses the full ID as a tiebreaker to ensure consistent output ordering across Python/pandas versions.

Expected test outputs were updated to reflect the new deterministic ordering.


susannasiebert (approved):
+1

PR #1420, opened 2026-07-13
Refactor prediction pipeline

susannasiebert:
This PR refactors the complete pipeline that power the pVACseq/fuse/splice/bind run commands.

The new workflow is controlled by a pipeline-specific RunPipeline each inheriting from the RunPipeline base class. This takes care of the following:
  - checking the inputs and converting inputs as necessary
  - calling a pipeline-specific InputToKmerPipeline that takes care of taking the input, creating a transcript fasta of mutant and (if applicable) wild type transcript sequences, and breaking up these transcript sequences into mutant and matched wildtype kmers depending on the selected epitope lengths 
  - Setting up arguments for making predictions or class I and/or class II
  - Calling a pipeline-specific PredictionPipeline which takes care of splitting the kmer fastas into smaller subsets based on the fasta-size, make prediction calls, parsing the outputs, combining the outputs into one all_epitopes file
  - Calling the existing PostProcessor
 A lot of the main logic was duplicated in each run.py file and was able to be abstracted and moved to the RunPipeline.

The existing CallIedb module was removed and replaced with a CallPredictors. Logic for splitting out the kmer fasta into chunks was moved from the old Pipeline module into here and combined with the old CallIedb logic for calling the selected predictors.

The OutputParser module has been cleaned up. Logic for matching up mutant and wild type kmers has been removed since that is now handled by the InputToKmerPipelines. Logic for normalizing percentiles has been moved to a new module NormalizedPercentileCalculator. Duplicate code has been eliminated and inheritance logic has been streamlined. The OutputParser now defines two child classes MatchedSequencesOutputParser and UnmatchedSequencesOutputParser which handle the majority of the logic for parsing the output files from the predictors, generating summary metrics, etc. Pipeline-specific output parsers inherit from these two classes, as appropriate, and only define the pipeline-specific output headers and constructing each row to be written to the output file.

The existing Pipeline and FataGenerator modules have been removed since they no longer used and all their logic is now handled in the new modules.

Test have been updated, moved to more appropriate test classes, or removed if no longer needs. There may be a few tests that were removed in the output parser tests that should be tested in the InputToKmerPipeline tests since they focused on the logic for matching mutant and wild type kmers. Those will be addressed in a future PR.

The predictor_tests have been updated to use a kmer fasta. I also updated the IEDB-specific tests in there to use standalone IEDB instead of the API or mocking the files.


pvactools/lib/call_predictors.py
```diff
@@ -0,0 +1,137 @@
...
+        self.n_threads = kwargs['n_threads']
+        self.output_dir = kwargs['output_dir']
+        self.tmp_dir = os.path.join(self.output_dir, 'tmp')
+        os.makedirs(self.tmp_dir, exist_ok=True)
+        self.log_dir = os.path.join(self.output_dir, 'log')
+        os.makedirs(self.log_dir, exist_ok=True),
+        self.output_files = []
+        self.output_key_files = []
+
+    def get_flurry_state(self):
```
tmooney: It's a little suspicious that something named `get` ends up causing changes to the prediction_algorithms.  As written, calling it twice wouldn't be safe--maybe it should directly return `self.flurry_state` if it's already set?

pvactools/lib/normalized_percentile_calculator.py
```diff
@@ -0,0 +1,140 @@
...
+                return True
+
+            if not event.is_set() and normalized_allele not in PercentileFileCache._missing:
+                pass
+            else:
+                # Another thread is already downloading – wait for it
+                event.wait()
+                return os.path.exists(file_path)
+
+        url = f"https://raw.githubusercontent.com/griffithlab/pvactools_percentiles_data/main/hdf5/{allele_file}"
```
tmooney, line 116: Is it safe to trust that the `main` version will always be an appropriate one to use here?  While my instinct is that pointing to versioned data is safer, there might be good reason to want to be able to update it (at the potential cost of reproducibility).

pvactools/lib/output_parser.py
```diff
@@ -1307,186 +690,80 @@ def add_prediction_scores(self, row, mt_scores):
...
-                mt_result['mutation_position'] = 'NA'
-        return iedb_results
+    def construct_row(self, result):
+        row = {
+            'HLA Allele'          : result['allele'],
+            'Sub-peptide Position': result['position'],
+            'Epitope Seq'         : result['mt_epitope_seq'],
+            'Index'               : result['tsv_index'],
+            #Median IC50 Score
+            'Median IC50 Score': self.rounded_score_or_na(result['median_mt_ic50']),
```
tmooney, line 785: I sort of enjoy how the comments in this section all act as call and response where the line immediately after it is exactly the thing the comment said it would be 🙂

pvactools/lib/pvacseq_run_pipeline.py
```diff
@@ -0,0 +1,86 @@
...
+        post_processing_params['filtered_report_file'] = filtered_file
+        PvacseqPostProcessor(**post_processing_params).execute()
+
+    def call_ml_predictor(self):
+        if len(self.class_i_prediction_algorithms) > 0 and len(self.class_i_alleles) > 0 and len(self.class_ii_prediction_algorithms) > 0 and len(self.class_ii_alleles) > 0:
+            # Run ML predictions
+            if self.run_ml_predictions:
+                logging.info("Running ML predictions...")
+                if not 'all' in self.prediction_algorithms:
+                    logging.info("Caution: Use 'all' in prediction_algorithms is strongly recommended. Missing features will be filled with NA and will cause predictions to be inaccurate. Running ML predictions regardless...")
```
tmooney:
```suggestion
                    logging.warning("Caution: Using 'all' in prediction_algorithms is strongly recommended. Missing features will be filled with NA and will cause predictions to be inaccurate. Running ML predictions regardless...")
```
There are other ways to make this grammatical as well, like "Use of". I might prefer "Including" even.  Side note--does this rise to the level of a `warning` or is `info` okay?

pvactools/lib/run_argument_parser.py
```diff
@@ -577,7 +574,7 @@ def __init__(self):
 class PvacvectorRunArgumentParser(RunArgumentParser):
     def __init__(self):
         tool_name = 'pvacvector'
-        input_file_help = "A .fa file with peptides or a pVACseq .tsv file with epitopes to use for vector design."
+        input_file_help = "A .fa file with peptides or to use for vector design."
```
tmooney: This help text has some awkward parallelism.  Is this `a .fa file` that is one of these two possibilities?
- `with peptides`
- `to use for vector design`

If so, maybe adding the word `either` before `with peptides` would make it more clear. If not, then I'm not sure what it's going for.

tests/test_normalized_percentile_calculator.py
```diff
@@ -0,0 +1,57 @@
...
+
+from pvactools.lib.normalized_percentile_calculator import NormalizedPercentileCalculator
+from tests.utils import *
+
+class OutputParserTests(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        executable_dir    = os.path.join(pvactools_directory(), 'pvactools', 'lib')
+        cls.executable    = os.path.join(executable_dir, 'normalized_percentile_calculator.py')
+        url = f"https://raw.githubusercontent.com/griffithlab/pvactools_percentiles_data/main/hdf5/HLA-A_02_01_percentiles.h5"
```
tmooney, line 14: Commenting here as a duplicate of my previous comment just to link them together.

tmooney (commented):
I tried to at least look at every code file in this diff, but it would be wildly overstating it to claim I thoroughly read and understood every detail.

pvactools/lib/call_predictors.py
susannasiebert: I honestly hate this whole flurry state logic. Using MHCflurry gives you presentation and binding data. This logic was implemented so that you could chose to use only one of the two but also use both without having to call MHCflurry twice. I kinda want to just remove the MHCflurryEL (presentation) option and have MHCflurry always give you both results. That would clean up the code in various places (including the output parser). It's something I will discuss with the team. In the meantime I agree that returning flurry_state here when it is already set is the right move. I will make that change.

pvactools/lib/normalized_percentile_calculator.py
susannasiebert: I think this is a good idea and something we should add in the future. For right now while this data might change it will only change to add new algorithms, alleles, or lengths, not change the existing data. But in the future, existing raw scores might be updated e.g. if the version of an underlying algorithm changes. 

ETA: I would want to do something more sophisticated like making tagged/branched versions of the scores repo and then explicitly updating them in concert with pVACtools releases.

pvactools/lib/output_parser.py
susannasiebert: I mostly just wanted to mirror what I did for the other tools-specific parsers :). The pVACbind logic is a lot simpler since it's fewer scores per type but in the other tools it gets confusing fast.

pvactools/lib/pvacseq_run_pipeline.py
susannasiebert: This was originally just a "print" statement so I thought replacing it with info-level logging would be fine. But it does seem more like  a warning sort of messages. I'll change it.

pvactools/lib/run_argument_parser.py
susannasiebert: Nope, this was just my oversight. We removed the option of starting with a tsv file so the `or` needs removing.

PR #1421, opened 2026-07-14
Move shared GenerateProteinFasta code to a base module

susannasiebert:
After the changes made in #1420 a lot of code was identical between the different generate protein fasta modules. This PR adds a new base class GenerateProteinFasta that controls this shared code. Tool-specific classes inherit from this base class and have additional code that differs between the tools (the code to generate the transcript fasta, and the trimming logic for these transcript fastas). 

As a side effect, pVACfuse and pVACsplice fasta generated with this command will now also contain the Best Peptide in the MT/ALT headers if ran with an aggregated report input tsv. This is a pre-req for enabling use of these outputs with pVACvector and the peptide ordering form creation.


pvactools/lib/generate_protein_fasta.py
```diff
@@ -0,0 +1,323 @@
...
+        self.pass_only = kwargs.pop('pass_only', False)
+        self.biotypes = kwargs.pop('biotypes', ['protein_coding'])
+        self.allow_incomplete_transcripts = kwargs.pop('allow_incomplete_transcripts', False)
+        self.flanking_sequence_length = kwargs.pop('flanking_sequence_length')
+        self.mutant_only = kwargs.pop('mutant_only', False)
+        self.aggregate_report_evaluation = kwargs.pop('aggregate_report_evaluation', ['Accept'])
+        self.temp_dir = tempfile.mkdtemp()
+        self.fasta_file_path = kwargs.pop('fasta_file_path', os.path.join(self.temp_dir, f"{self.sample_name}.transcripts.fa"))
+        self.trimmed_fasta_file_path = kwargs.pop('trimmed_fasta_file_path', os.path.join(self.temp_dir, f"{self.sample_name}.transcripts.trimmed.fa"))
+        self.filtered_fasta_file_path = os.path.join(self.temp_dir, f"{self.sample_name}.transcripts.filtered.fa")
```
tmooney, line 28: Still scary to put sample names right into filenames 🙂

pvactools/lib/generate_protein_fasta.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def generate_fasta(self):
+        raise Exception("Implement in child class")
+
+    def trim_sequences(self):
+        raise Exception("Implement in child class")
+
+    def filter_fasta(self):
+        raise Exception("Implement in child class")
+
```
tmooney:
```suggestion
```

It looks like `filter_fasta` is defined down below already and neither child class implements it.

susannasiebert: Good catch. The filter_fasta methods used to be defined in the child classes when I first moved stuff around before I normalized this method.

pvactools/lib/generate_protein_fasta.py
susannasiebert: Yeah, understandable. It's a paradigm we use throughout pVACtools for output file names so it follows that.

tmooney (approved):
+1 🦖

PR #1422, opened 2026-07-15
Harmonize downstream sequence length parameter between pVACseq/fuse/splice

susannasiebert:
This PR adds the downstream sequence length parameter option to pVACsplice and fixes how it is used in pVACfuse to make the behaviors consistent between pVACseq, fuse, and splice. It also adds tests for this new parameter to the JunctionToFasta and FusionToFasta tests.

It also adds a new argument parser type that handles parsing this parameter.

It moves all argument parser type definition out of run_utils.py to a new run_argument_utils.py file.

Lastly it fixes a bug in list of tiers available in the tiers argument parser to update them to the current list of tiers.


tmooney (approved):
+1, nice simplifying a bunch of the imports!

PR #1424, opened 2026-07-15
Remove code in predict() methods that split sequences into kmers

susannasiebert:
Because the pipeline are now creating kmer fastas and feed those to the predictors there is no need for the predictors to split the peptides into kmers.


tmooney (approved):
+1 🧥

PR #1426, opened 2026-07-16
Speed up unit tests

susannasiebert:
IO on GitHub action runners is exceedingly slow. This especially affected various pVACsplice tests which rely on a GTF reference. These, even when pared down to only a single chromosome, can still be hundreds of thousands of entries. 

This PR replaces all gtf files used in the test with ones that only contain entries for genes in the result set or needed for testing specific functionality (e.g. cds_start_NF, cd_end_NF flags to test the allow incomplete transcripts option).

These files are now also saved in raw format, not gzipped, because unzipping them (or having the gtfparse library handling unzipping) is another bottleneck.

In order to ascertain test runtimes the previous execution via unittest has been replaced by pytest (which has unittest support and provides a flag for outputting test runtime information).

Runtime for unittests is now down to ~20 minutes from over 2 hours.

Side note: the slowness of gtf parsing seems to be specific to GitHub actions as these tests were already running at more reasonable speeds locally.


tmooney (approved):
+1, seems good to me. Kinda goofy that Github is so slow with the other testcases!

PR #1427, opened 2026-07-17
Update MHCflurry to run/parse processing and presentation scores

susannasiebert:
This PR removes the `MHCflurryEL` prediction algorithm option. Instead when a user selects `MHCflurry` it will now parse the presentation and processing scores in addition to the ic50 scores. This removes the need for the flurry state to keep track of which data needs to be parsed from the output file.


tmooney (approved):
+1, yay removing the flurry_state ❄️

PR #1428, opened 2026-07-17
Clean up calculate reference proteome module

susannasiebert:
With the refactor of the pipeline processing, the pipelines now produce a fasta file where the sequences are already trimmed to 7 flanking amino acids around the mutation and feed that fasta into the calculate reference proteome similarity step. These sequences can now be directly used to query blast / the reference proteome fasta because at a match length of 8 there won't be any false-positive matches.

The standalone calculate reference proteome similarity steps now expect a fasta file with a 7 flanking sequence length.

This PR removes a lot of unneeded code that was handling false positive matches, which are no longer needed.


pvactools/lib/calculate_reference_proteome_similarity.py
```diff
@@ -285,38 +217,17 @@ def _get_peptide(self, line, mt_records_dict, wt_records_dict):
...
-                return None, None, None
-
-            # get peptide
-            subpeptide_position = full_peptide.index(epitope)
-            if variant_type == 'FS':
-                peptide = self.extract_n_mer_from_fs(full_peptide, wt_peptide, epitope, subpeptide_position)
-            else:
-                peptide = self.extract_n_mer(full_peptide, wt_peptide)
-        return peptide, full_peptide, wt_peptide
+            raise Exception("Unsupported file type")
```
tmooney: Might be nice if this exception included info on what type it got.

pvactools/lib/calculate_reference_proteome_similarity.py
```diff
@@ -623,7 +529,8 @@ def parser(cls, tool):
             'input_fasta',
             help="For pVACbind, the original input FASTA file. "
             + "For pVACseq, pVACfuse, and pVACsplice a FASTA file with wildtype and mutant peptide sequences for each variant isoform with 7 amino acids flanking sequence. "
-            + "This file can be found in the same directory as the input filtered.tsv/all_epitopes.tsv file. "
+            + "This file can be found in the same directory as the input filtered.tsv/all_epitopes/aggregated tsv file. "
```
tmooney: Should this be `aggregated.tsv` file?

tmooney (approved):
+1 🪂

PR #1429, opened 2026-07-17
Fix a couple of issues with NetChop after pipeline refactor

susannasiebert:
This PR addresses two issues with NetChop that arose after the pipeline refactor:

- The current parsing logic would include some positions around the actual epitope in the list of cleavage sites
- The fasta that would be passed into the NetChop step would not include enough amino acids. NetChop requires a flanking sequence of 9 amino acids around the epitope (not mutation) in order to provide consistent cleavage scores. Updating the flanking sequence length to 9 + max(epitope_lengths) ensures that there are always sufficient flanking amino acids around all possible epitopes, no matter where they fall in relation to the mutation.

This PR also updates mock and test data to match current inputs.


tmooney (approved):
+1 🪓

PR #1432, opened 2026-07-20
Fix NetChop bug that would include position before epitope (0)

susannasiebert:
This bug was first discovered as part of #1429 and this PR incorporates the bugfix into the staging branch for the upcoming 7.1 release.


tmooney (approved):
+1 🦓

PR #1430, opened 2026-07-18
Harden filtering and runtime error handling

YT-er:
## Summary

- replace dynamic `eval()` calls in filtering and transcript prioritization with explicit operator and boolean parsing
- preserve command-level `AttributeError` exceptions instead of treating them as missing CLI commands
- replace a production debugger breakpoint in missense output parsing with a contextual error
- add regression coverage for malicious input, CLI error propagation, and missense matching failures

## Why

Several runtime paths relied on dynamic evaluation or overly broad exception handling. This could execute untrusted TSV content, mask real command failures, or leave production runs in a debugger.

## Impact

Valid inputs retain their existing behavior while invalid filter criteria and transcript flags now fail deterministically with actionable errors. Unexpected command failures propagate to callers, and inconsistent missense output now reports the missing position and available context.

## Validation

- 23 focused regression tests for the changes in this PR passed
- 28 existing output parser regression tests passed
- 3 targeted integration tests for the fixes merged in #1433 and #1434 passed after rebasing
- Python compile check passed for `pvactools` and `tests`
- `pip check` reported no broken requirements
- diff whitespace check passed

## Follow-up

This branch was rebased onto `upstream/master` at v7.1.1. Changes already merged through #1433 and #1434 were removed from this PR.

PR #1433, opened 2026-07-20
Fix bug in lowexpr logic

susannasiebert:
This PR incorporates the bugfix regarding the LowExpr tier logic first discovered in #1430. This PR separates it out so that it can be incorporated separately from the other changes proposed in #1430.


tmooney (approved):
+1, I suppose that `['RNA Depth']` has always been not equal to `'NA'`!

PR #1434, opened 2026-07-20
Fix tbi file existence check in pVACsplice runs

susannasiebert:
This PR incorporates the bugfix regarding an incorrect file existence check first discovered in https://github.com/griffithlab/pVACtools/pull/1430. This PR separates it out so that it can be incorporated separately from the other changes proposed in https://github.com/griffithlab/pVACtools/pull/1430.


tmooney (approved):
+1 🪞

PR #1436, opened 2026-07-20
Fix TLBind WT Presentation Percentile not getting set correctly

susannasiebert:
This PR fixes an issue where - for the `TLBind WT Presentation Percentile` output field - the logic was not using the correct key to pull the value from the data structure.


tmooney (approved):
+1 🧗

PR #1438, opened 2026-07-22
Re-enable deletion of tmp files unless -k flag is set

susannasiebert:
After refactoring the prediction pipeline tmp files were no longer getting deleted. This PR re-enables that behavior.

### DIFF
--- a/pvactools/lib/run_pipeline.py
+++ b/pvactools/lib/run_pipeline.py
@@ -2,6 +2,7 @@ import sys
 import os
 import platform
 import logging
+import shutil
 
 from pvactools.lib.prediction_class import NetMHCIIVersion
 from pvactools.lib.print_log import print_log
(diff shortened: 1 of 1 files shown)


tmooney (approved):
+1, I do wonder about this particular method needing to know so many details of constructing the paths in order to clean them up.

susannasiebert:
I had debated putting this at the end of [prediction_pipeline.py](https://github.com/griffithlab/pVACtools/blob/8.0.0/pvactools/lib/prediction_pipeline.py). There we have the `per_epitope_output_dir`, which will have the `tmp` directory underneath it. Do you think that would be the better spot for it?

tmooney:
Interesting! I guess that would clean up each one as it goes along instead of saving them all for the end?  Could be nice for saving space as long as it wouldn't end up removing something that someone debugging might want to look at if a different step fails.  I would understand the desire to save it all until the end, in which case where it is now is preferable.  (Unless we can have something that we can callback to clean up, but that's a lot more complex a solution...)

susannasiebert:
Yeah, that was my main concern of saving it to for the very end.

PR #1439, opened 2026-07-22
Restore several output parser tests and move to variant_to_fasta/fasta_to_kmer tests

susannasiebert:
Various output parser tests before the pipeline refactor where testing the matching up of wildtype to mutant peptides. Now that that logic is handled in the fasta to kmers step, this PR adds these tests to the variant to fasta and fasta to kmer tests.

This PR also fixes an issue where inframe insertions where not handled correctly.


tests/test_output_parser.py
```diff
@@ -89,8 +87,6 @@ def test_parse_output_runs_and_produces_expected_output_with_all_class_i_files_n
 
         self.assertFalse(parser.execute())
         expected_output_file  = os.path.join(self.test_data_dir, "output.all_class_i.normalized_percentiles.tsv")
-        import shutil
-        shutil.copy(output_file.name, expected_output_file)
```
tmooney, line 93: This would make it easier for the files to match, wouldn't it? 😄

tests/test_variant_to_fasta.py
```diff
@@ -509,5 +509,85 @@ def test_input_file_with_protein_altering_variant_insertion_generates_expected_f
...
+
+        generate_fasta_params = {
+            'input_file'                : generate_fasta_input_file,
+            'output_file'               : generate_fasta_output_file.name,
+            'downstream_sequence_length': None,
+            'proximal_variants_file'    : None,
+        }
+        generator = VariantToFasta(**generate_fasta_params)
+
+        self.assertFalse(generator.execute())
```
tmooney: I realize that's how all these tests are written, but it seems weird to me to assertFalse for an execution.  From looking at the code, it seems like `execute` has no explicit return value so I'm wondering in what cirumstances this test might fail?

tmooney (approved):
+1 (as usual I am not reviewing the various test inputs/outputs so hopefully all the FASTAs and the like are you like them!)

tests/test_variant_to_fasta.py
susannasiebert: You're totally right. I at some point got it into my head that this was the correct way to check that a method executed successfully but doing some more research both unittest and pytest will fail a test if a method that is called as part of the test throws an error. I pushed a commit to remove these checks at least from the test files touched here.

PR #1440, opened 2026-07-23
Fix gene of interest highlighting

susannasiebert:
Gene of interest highlighting was not working consistently in pVACview 8.0. This PR fixes that.


tmooney (approved):
+1, I think this makes sense! (I'm a lot less familiar with the R code in this project.)

PR #1441, opened 2026-07-27
 Add create_peptide_ordering_form command to pVACfuse and pVACsplice

susannasiebert:
This PR moves the central logic for creating the peptide ordering form to pvactools/lib. It updates the logic to handle files from the pVACfuse and pVACsplice pipelines and adds standalone commands to those tools' command line tree.

It also fixes a couple of bugs:
- The existing way of generating the peptide fastas as part of creating the peptide ordering form was not correct and would not create a fasta of mt and wt sequences. The generate peptide fasta logic was updated to support this use case.
- the sample name was not being passed into FusionToKmerPipeline step when running the generate protein fasta command
- The prediction file would throw an error when there were no result files from the predictor. This can happen when a given allele-length-predictors combo is invalid, i.e. the allele, length, or the allele-length combo is not supported by any of the selected predictors.


pvactools/lib/create_peptide_ordering_form.py
```diff
@@ -0,0 +1,354 @@
...
+            )
+        elif tool == 'pvacsplice':
+            parser.add_argument(
+                "input_file",
+                help="RegTools junctions output TSV file"
+            )
+            parser.add_argument(
+                "annotated_vcf",
+                help="A VEP-annotated single- or multi-sample VCF containing genotype and transcript information."
+                + "The VCF ma be gzipped (requires tabix index)."
```
tmooney:
```suggestion
                + "The VCF may be gzipped (requires tabix index)."
```

tmooney (approved):
+1 this generally makes sense to me!

PR #1446, opened 2026-07-30
Clean up packaging metadata and repository artifacts

YT-er:
## Summary

- Remove seven tracked macOS `.DS_Store` files and exclude them from future distribution manifests.
- Limit namespace package discovery to `pvactools` and its subpackages.
- Add the supported Python range to package metadata.
- Update the README and installation guide to match the existing Python 3.9–3.11 runtime support.

## Motivation and impact

Unrestricted namespace package discovery treated top-level repository directories such as `tests` and `docs` as installable packages. The package metadata also lacked `Requires-Python`, so package installers could not reject unsupported Python versions during dependency resolution. In addition, tracked `.DS_Store` files could be included in release artifacts, while the installation documentation still advertised Python 3.6 or newer.

This change keeps installed wheels scoped to `pvactools`, exposes the existing Python support contract to package installers, aligns the documentation, and removes macOS metadata from the repository and release manifests.

## Checks

- Package discovery returns 158 packages, all under the `pvactools` namespace.
- Generated egg-info contains `Requires-Python: >=3.9,<3.12` and a single `pvactools` top-level package.
- Built wheel contains only `pvactools` and dist-info entries, with no `.DS_Store`, `tests`, or `docs` top-level packages.
- Installed the wheel into a temporary target and imported `pvactools.tools.main` from that installation.
- `python -m unittest tests.test_tools_main`
- `git diff --check`


susannasiebert (approved):
+1

PR #1452, opened 2026-08-12
Implement new select, select_class_i, and select_class_ii algorithm options to run subset of algorithms

### LINKED ISSUE #1449: Implement a "recommended" algorithm option
Now that we have added a lot of new algorithms, running with the `all` option can be very time consuming. We should add a `recommended` option that only runs a subset of the algorithms.

A few thoughts on algorithms to pick:
- Between SMM and SMMPMBEC, decide on one. They're highly correlated.
- Decide whether to exclude MHCflurry in favor of BigMHC.
- Pick one of the NetMHC algorithms. NetMHCpan seems to be popular.
- Exclude TL-MHC algorithms because they are slow.

susannasiebert:
Closes #1449

This PR also fixes a few typos in the docs.

### DIFF
--- a/docs/pvacbind/features.rst
+++ b/docs/pvacbind/features.rst
@@ -75,7 +75,7 @@ for a variety of additional algorithms.
      - Immunogenicity
      - no
      - yes
-   * - ImmuoScope_IM
+   * - ImmuScope_IM
      -
      - MHC Class II
      - Immunogenicity
@@ -165,7 +165,7 @@ for a variety of additional algorithms.
      - Immunogenicity
      - yes
      - yes
-   * - Pickpocket
+   * - PickPocket
      - 1.1
      - MHC Class I
      - Binding
--- a/docs/pvacbind/features.rst
+++ b/docs/pvacbind/features.rst
@@ -46,7 +46,11 @@ and immunogenicity algorithms.
 By using the IEDB RESTful web interface, pVACbind leverages their extensive support of different prediction algorithms.
 
 In addition to IEDB-supported prediction algorithms, we've also added support
-for a variety of additional algorithms.
+for a variety of additional algorithms. The following table lists all algorithms supported
+by pVACbind, whether the algorithm is a MHC class I or class II algorithm, what type of scoring
+the algorithm returns (binding, presentation, or immunogenicity), whether it supports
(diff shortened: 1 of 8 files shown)


docs/pvacbind/features.rst
```diff
@@ -46,7 +46,11 @@ and immunogenicity algorithms.
 By using the IEDB RESTful web interface, pVACbind leverages their extensive support of different prediction algorithms.
 
 In addition to IEDB-supported prediction algorithms, we've also added support
-for a variety of additional algorithms.
+for a variety of additional algorithms. The following table lists all algorithms supported
+by pVACbind, whether the algorithm is a MHC class I or class II algorithm, what type of scoring
+the algorithm returns (binding, presentation, or immunogenicity), whether it supports
+percentile ranks natively, supports normalized percentile ranks (``--use-normalized-percentiles``
+flag), and whether the ``select`` algorithm shortcut includes this algorithm.
```
tmooney: I kind of think the parallelism of these descriptions could be strengthened.  There's a bit of inconsistency in how they're worded.  As an example alternative: 
```suggestion
for a variety of additional algorithms. The following table lists all algorithms supported
by pVACbind, whether the algorithm is a MHC class I or class II algorithm, what type of scoring
the algorithm returns (binding, presentation, or immunogenicity), whether the algorithm supports
percentile ranks natively, whether the algorithm supports normalized percentile ranks (``--use-normalized-percentiles
flag), and whether the ``select`` algorithm shortcut includes this algorithm.
```

I kind of want to replace one of the two words "algorithm" in the MHC Class description, too.

tmooney (approved):
+1 🍞

docs/pvacbind/features.rst
susannasiebert: What about this?


```suggestion
for a variety of additional algorithms. The following table lists all prediction algorithms supported
by pVACbind, whether the predictor is a MHC class I or class II algorithm, what type of scoring
the predictor returns (binding, presentation, or immunogenicity), whether the algorithm supports
percentile ranks natively, whether the predictor supports normalized percentile ranks (``--use-normalized-percentiles
flag), and whether the ``select`` algorithm shortcut includes this predictor.
```

PR #1454, opened 2026-08-24
Normalize MixMHC2pred alleles

susannasiebert:
The previous list of MixMHC2pred alleles (`OfficialName` column of the alleles file) had the `DP` and `DQ` alpha and beta chain pairs linked with an `/` instead of a `-`. This had two problems:
1) When someone would specify the normal format (with an `-`, e.g. `DPA1*01:03-DPB1*02:01.15`), it would skip MixMHC2pred because this allele wasn't matching the expected MixMHC2pred format.
2) When specifying the allele with the expected MixMHC2pred format (e.g. `DPA1*01:03/DPB1*02:01.15`) it would throw an error when trying to save the results because the result file has the allele in its name and the forward slash would be interpreted as a directory path

Updating the MixMHC2pred allele list to use `-` instead of `/` fixes this problem. No other changes are needed because the MixMHC2pred prediction code already looks up a custom MixMHC2pred formatting for each allele (`AlleleName` column of the alleles file).


tmooney (approved):
+1 for this change, but also the thought of putting `*` into filenames is not thrilling me either 🙂.

Would it be reasonable to have some version of a sanitizer like the old https://github.com/genome/genome/blob/d9bdd38ffbf32be2ea295629e4587a9b343d6d69/lib/perl/Genome/Utility/Text.pm#L125-L134 ?  (That exact implementation can lead to collisions if someone has pathological names...).  Adding one would be way beyond the scope of this PR, of course!

susannasiebert:
I think this would be outside of the scope of this PR but I'll make an issue.

PR #1456, opened 2026-08-25
Correctly handle gene name parsing from fasta headers for genes with lowercase characters

susannasiebert:
Without including lowercase characters in the regex, some gene names would not be parsed correctly. E.g. `Cacna1e` would be shortened to `C`. This fix is tested in a new test case (`test_calculate_self_similarity_with_aggregated_tsv_and_peptide_fasta_mouse_with_matches`) and can be confirmed by looking at the `tests/test_data/calculate_reference_proteome_similarity/output_mouse.all_epitopes.aggregated.metrics.json` output file which now has the entry `"Genes": "Cacna1e",` in the `reference_matches` key, which would previously only be `"Genes": "C",`.


pvactools/lib/calculate_reference_proteome_similarity.py
```diff
@@ -541,7 +541,7 @@ def _write_outputs(self, processed_peptides, mt_records_dict, wt_records_dict):
                         matches = []
                         for query_window, hit_reference_matches in groupby(metric_lines,key=lambda x:x['Match Window']):
                             hit_reference_matches = list(hit_reference_matches)
-                            gene_regex = '^.*gene_symbol:([0-9|A-Z]+).*$'
+                            gene_regex = '^.*gene_symbol:([0-9|A-Z|a-z]+).*$'
                             transcript_regex = '^.*transcript:(ENS[0-9|A-Z|.]+).*$'
```
tmooney, line 545: Do we need to worry about lowercase letters here, too?  On [this page](https://jun2026.archive.ensembl.org/info/genome/stable_ids/prefixes.html) I see one with lower-case letters (and a couple with `_`s), but it feels like quite an edge case!

tmooney (approved):
+1 🌻

pvactools/lib/calculate_reference_proteome_similarity.py
susannasiebert: Because pVACtools only supports a limited number of species, and none of them use lowercase characters or underscores in their identifier, I think we're good on that front.

PR #1458, opened 2026-08-31
Fix command mentions of the allele_specific_cutoffs command

susannasiebert:
In various places we were still referencing the previous command line tree of the `allele_specific_cutoffs` command (`pvacseq|fuse|splice|bind allele_specific_cutoffs` instead of `pvactools allele_specific_cutoffs`). This PR fixes that.


tmooney (commented):
How about this one? https://github.com/griffithlab/pVACtools/blob/af8487df47127f4c6640203cdaa1e11b8b6d5934/pvactools/lib/update_tiers.py#L79

susannasiebert:
> How about this one?
> 
> https://github.com/griffithlab/pVACtools/blob/af8487df47127f4c6640203cdaa1e11b8b6d5934/pvactools/lib/update_tiers.py#L79

Good catch. Fixed.

tmooney (approved):
+1 📟

PR #1459, opened 2026-09-01
Fix issue in the Transcript Pass logic that would cause pVACview to crash on retiering

susannasiebert:
Using `apply(df$mainTable, TRUE ...` instead of `apply(df$mainTable, 1 ...` when calculating `transcript_pass` status would cause pVACview to crash during re-tiering. This PR fixes that. It also centralizes the logic for this status by re-using the existing `is_transcript_pass` function. Additionally, the exiting logic to add a new column for the results of this function during retiering would result in a second Transcript Pass column to be appended instead of the existing column to be updated. This PR fixes that.


tmooney (approved):
+1 🏗️

PR #1460, opened 2026-09-04
Add new argument to aggregate report creation to limit predictions to specific alleles

susannasiebert:
This PR adds a new functionality to the `pvacseq|bind|fuse|splice generate_aggregated_report` commands. A new parameter `--limiting-alleles` allows user to specific a set of alleles that are part of alleles in the input file and only process those entries.


pvactools/lib/aggregate_all_epitopes.py
```diff
@@ -20,6 +20,10 @@
 class AggregateAllEpitopes:
     def __init__(self):
         self.hla_types = pd.read_csv(self.input_file, delimiter="\t", usecols=["HLA Allele"])['HLA Allele'].unique()
+        if self.limiting_alleles:
+            extra_alleles = list(set(self.limiting_alleles) - set(self.hla_types))
+            if len(extra_alleles) > 0:
+                raise Exception(f"Alleles specified in the --alleles parameter not found in the input file: {extra_alleles.join(', ')}")
```
tmooney, line 26: Pedantically I'd say that `Allele(s)` or `One or more alleles` is more correct.

pvactools/tools/pvacbind/generate_aggregated_report.py
```diff
@@ -83,6 +83,11 @@ def define_parser():
              + "Whether the lowest or median is considered for each metric is controlled by the --top-score-metric parameter. ",
         default=['ic50', 'combined_percentile'],
     )
+    parser.add_argument(
+        "--limiting-alleles", type=lambda s:[a for a in s.split(',')],
+        help="Comma-separated list of alleles used for the predictions made in the input file. "
+             + "If specified, only predictions for those alleles will be included in the aggregated report."
```
tmooney, line 89:
```suggestion
             + "If specified, only predictions for these alleles will be included in the aggregated report."
```
I'd say this applies to `these` ones here and not `those` other ones there. 🙂(I'll not repeat this comment the other three places this one appears.)

tmooney (approved):
+1, looks fine to me!  It's clear what's happening, but `df = df[df[...` is kind of entertaining out of context 🙂

PR #1462, opened 2026-09-08
Respect IMMU_SCOPE_WEIGHTS_DIR as first priority for determining ImmuScope weight directory

### LINKED ISSUE #1461: Support custom weights directory in immuscope-download-weights
Hello! Thank you for creating this tool!

I wanted to install pVACtools and ImmuScope in a Docker image and ran into a small issue with the model weights installation.

When I run:

```Dockerfile
RUN pip install git+https://github.com/griffithlab/ImmuScope.git#egg=ImmuScope && immuscope-download-weights
```
the weights are installed in `/root/.local/share/ImmuScope/weights`, because `XDG_DATA_HOME` is not set. This becomes an issue when the container is later run as a non-root user, since the user will not have access to `/root/.local/share/ImmuScope/weights`.

I noticed that `immuscope_wrapper.py` supports overriding the weights location via the `IMMU_SCOPE_WEIGHTS_DIR` environment variable, but `download_weights.py` currently only uses `XDG_DATA_HOME` to determine the installation directory.

Would it be possible to add support for `IMMU_SCOPE_WEIGHTS_DIR` to `download_weights.py` as well, using the same logic as in `immuscope_wrapper.py`?

For now I use the following solution:
```Dockerfile
RUN pip install git+https://github.com/griffithlab/ImmuScope.git#egg=ImmuScope && XDG_DATA_HOME="$SOFT" && immuscope-download-weights
```

Thanks for your time!

susannasiebert:
Closes #1461 

This PR needs to be merged in conjunction with https://github.com/griffithlab/ImmuScope/pull/1.

Currently, the logic in `resolved_immuscope_weights_dir` doesn't respect/take into account `IMMU_SCOPE_WEIGHTS_DIR`. Instead only `XDG_DATA_HOME` and the default `~/.local/share'/ImmuScope/weights` are used. This PR updates the logic to first look at `IMMU_SCOPE_WEIGHTS_DIR`. `IMMU_SCOPE_WEIGHTS_DIR` is already the first priority for the weights directory when running the ImmuScope predictor via the `immuscope_wrapper.py `. After merging in https://github.com/griffithlab/ImmuScope/pull/1, this env variable will also be the first one to be used for downloading the weights. 

With this change in these two PRs the `IMMU_SCOPE_WEIGHTS_DIR` will now be consistently respected for downloading the weights and running ImmuScope inside of pVACtools.

### DIFF
--- a/pvactools/lib/prediction_class.py
+++ b/pvactools/lib/prediction_class.py
@@ -942,19 +942,16 @@ class NetMHCIIVersion:
 
 class ImmuScope_IM(MHCII):
     immuscope_score_col = 'ImmuScope_IM'
-    immuscope_weights_dir = '/opt/ImmuScope/weights'
 
     def resolved_immuscope_weights_dir(self):
+        immu_scope_weights_dir = os.environ.get('IMMU_SCOPE_WEIGHTS_DIR')
         xdg_data_home = os.environ.get('XDG_DATA_HOME')
-        if xdg_data_home:
-            user_dir = os.path.join(xdg_data_home, 'ImmuScope', 'weights')
+        if immu_scope_weights_dir:
+            return immu_scope_weights_dir
+        elif xdg_data_home:
+            return os.path.join(xdg_data_home, 'ImmuScope', 'weights')
         else:
-            user_dir = os.path.join(os.path.expanduser('~/.local/share'), 'ImmuScope', 'weights')
-
-        if os.path.isdir(os.path.join(user_dir, 'IM')):
-            return user_dir
-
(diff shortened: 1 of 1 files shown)


tmooney (approved):
+1 as long as it's intentional to remove the possibility that `/opt/ImmuScope/weights` could be used.

susannasiebert:
I think that was shoe-horned in to already have a pre-set for our docker containers. I think it's fine because if someone wants to use that directory they can just set the env variable to point to it. I will make an addition to the docs to document this env variable as well.

PR #1463, opened 2026-09-09
Fix pVACview aggregate report header tooltips

susannasiebert:
After the changes made in v7.x to the aggregated report, we never updates the header tooltips when displaying this file in pVACview. This PR fixes that. It also fixes a few existing tooltips to better describe the contents of the column.


pvactools/tools/pvacview/styling.R
```diff
@@ -14,20 +14,27 @@ callback <- function(hla_count, score_mode) {
...
+    "        'Prob Pos - Problematic positions within the Best Peptide.',",
     "        'Num Included Peptides - The number of top-scoring, unique peptides included for review.',",
     "        'Num Passing Peptides - The number of unique well-binding peptides for this mutation.',",
-    paste("      'IC50 MT -", score_mode, "IC50 binding affinity of the best-binding mutant epitope across all prediction algorithms used.', "),
-    "        'IC50 WT - IC50 binding affinity of the corresponding wildtype epitope.',",
-    paste("      '%ile MT - ", score_mode, "binding affinity percentile rank of the best-binding mutant epitope across all prediction algorithms used (those that provide percentile output).', "),
-    "        '%ile WT - Binding affinity percentile rank of the corresponding wildtype epitope across all prediction algorithms used (those that provide percentile output).', ",
+    paste("      'IC50 MT - ", score_mode, "IC50 binding affinity of the Best Peptide across all binding affinity prediction algorithms used.', "),
```
tmooney, line 28: It looks like these `paste` lines have a different number of leading spaces from the literal strings in this list.  Is this intentional?

tmooney (approved):
+1, I'm trusting that these are now the correct columns 🙂

pvactools/tools/pvacview/styling.R
susannasiebert: I tested this change locally and it runs without issue so I don't think the number of spaces matters. I believe this is mainly done for human readability (I didn't write this originally so I'm not sure). I think the original intend was to maybe align all array entries with the first entry on line 14 but that no longer seems to be the case either way.

PR #1469, opened 2026-09-22
Set prime_allele default

susannasiebert:


tmooney (approved):
+1 good to have a value in the else case!

